PR #1, opened 2015-12-15
sql: always use "DROP TABLE IF EXISTS"

vincentbernat:
For both PostgreSQL and SQLite (already done for MySQL), this avoids an
error when creating the database the first time. A user may think there
was a problem during creation.

BTW, good to see pmacct on GitHub! :)


paololucente:
Hi Vincent,

Thanks a lot for the contribution, just merged :)

Cheers,
Paolo

PR #3, opened 2015-12-20
BGP starvation avoidance hack

FvDxxx:
Under heavy BGP update activities, later established BGP sessions might
starve as the first established ones get served first and thus even
keep-alives might not be handled for the later ones, causing them to
drop (and later reestablished). This hack avoids this.


paololucente:
Hi Markus,

Thanks a lot for your contribution!

Cheers,
Paolo

PR #7, opened 2016-02-23
Compatibility with Jansson 2.2

vincentbernat:
Hi Paolo!

I wanted to make pmacct compile on Precise which only ships with Jansson 2.2. It doesn't feature `json_object_update_missing()` function. Like for `strlcpy()`, I test for the presence of such a function and include its definition when absent.

The second commit also makes Travis use it.

I didn't run `autoreconf -i` as I have a widely different version of autotools than you have and this would pull so many changes. Moreover, for some reasons left to be investigated, this doesn't compile as expected unless I run the `bin/fix_autotools_timestamps.sh` but in this case, the `.deps` directory doesn't get created and the compilation fails. Before trying to investigate more, maybe you have some hints? I also notice that there are some missing `Makefile.am` in `src/*` directories. (insert an emoticon of someone being perplex)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ sudo: required
 
 install:
   - sudo apt-get install libtool autoconf cmake
-  - git clone https://github.com/akheron/jansson
-  - cd jansson && autoreconf -i && ./configure && make && sudo make install && cd ..
   - git clone https://github.com/edenhill/librdkafka
   - cd librdkafka && ./configure && make && sudo make install && cd ..
   - git clone https://github.com/alanxz/rabbitmq-c
@@ -27,4 +25,5 @@ addons:
       - libpq-dev
       - libsqlite3-dev
       - libmysqlclient-dev
+      - libjansson-dev
       - mysql-client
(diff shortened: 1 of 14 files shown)


job:
On Tue, Feb 23, 2016 at 08:08:08AM -0800, Vincent Bernat wrote:
> I wanted to make pmacct compile on Precise which only ships with
> Jansson 2.2. It doesn't feature `json_object_update_missing()`
> function. Like for `strlcpy()`, I test for the presence of such a
> function and include its definition when absent.

I have to ask, what is the benefit of supporting such an old version of
jasson? To run pmacct on Ubuntu 12.04 with jasson, what's wrong with
this strategy? https://github.com/pmacct/pmacct/blob/master/.travis.yml#L7-L8

Kind regards,

Job

vincentbernat:
It depends on people, but some people may dislike:

 1. having development tools on a server
 1. installing a library outside of the package management (no security fixes for example, less traceability)
 1. breaking other software that are installed and compiled against the older version of libjansson (but maybe the ABI is still backward compatible, so it may not apply in this case)

And I am one of them. If a more recent version of libjansson was a hard requirement, I would backport it for Precise (as a package) but in this case, it's easy to support an older version and it would benefit to other people like me who dislike going out their package management system for libraries.

job:
Thanks for your elaboration!

PR #9, opened 2016-02-24
* build: use more modern form for AM_INIT_AUTOMAKE

vincentbernat:
With recent automake versions, the macro AM_INIT_AUTOMAKE only takes a
list of flags as first argument. The version and project name should be
specified in AC_INIT and this was already the case. Moreover, we specify
we are not a GNU project (so we don't require some mandatory files).

**I didn't regenerate all files!**

I think it would be better to provide an "autogen" script and to remove autogenerated files from git. Released tarballs would still be able to include the generated files. I can do a PR to demonstrate this.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -2,7 +2,7 @@ dnl Process this file with autoconf to produce a configure script.
 dnl configuration file for pmacct 
 
 AC_INIT([src/pmacctd.c], [1.6.0-git], [paolo@pmacct.net])
-AM_INIT_AUTOMAKE([pmacctd], [1.6.0-git])
+AM_INIT_AUTOMAKE([foreign])
 AC_PREFIX_DEFAULT([/usr/local])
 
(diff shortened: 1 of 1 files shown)


vincentbernat:
The ancient automake currently used is also a bit buggy. If there are quotes in some configure arguments (for example `./configure '--mandir=${prefix}/share/man'`, the dependency tracking used by automake 1.4 will choke and bail out due to `echo '$(COMPILE) ...'` eating the first quote. This is fixed in more recent versions of automake. This explains why I was unable to compile in #7.

So, again, I can do a PR to remove autogenerated files and provide an `./autogen.sh` script (or just instructions to run `autoreconf -i`).

job:
I look forward to your PR

vincentbernat:
Ok, it's #10. Since it contains this one, I just close it.

PR #10, opened 2016-02-24
Improvements to the build system

vincentbernat:
Hey!

So, the changes can be picked individually if needed and I can argument on each of them.

The first change is just here to fix "autoreconf -i" when using a more modern automake. It should work just fine with older ones too. This is quite an old change. Dunno how configure and friends was generated without that.

The second change is to remove autogenerated changes and to provide an "./autogen.sh" script to generate them.

The third one moves the macro into m4. It's the usual location for them. Most projects don't use m4 anymore. It is also a stealth change to enable libtool that will be used for other changes.

The fourth change is to use libtool for plugins instead of building static libraries manually. Libtool can do convenience libraries. It's really just a bunch of .o and the information for linking together. It's not a real library. It's more flexible than static libraries because you keep the information about what should be linked together, so you don't need to use `-lsomething` when building the final program.

The fifth change is to ensure that make dist work. This way, you can get the release tarball with `make dist`. This is an invading change and I am using convenience libraries to minimize the amount of boilerplates. Automake needs to know all files. So, files cannot be added to a variable in `./configure` (automake needs to know unused files too) and we need to add all the `.h`.

The sixth change is to use silent rules. I find them more pretty.

Note that convenience libraries could be replaced by automake subdir-objects. It just means that we use variables and we can refer to objects in other directories (something not allowed in normal operations). We lose the ability to add libs to the convenience library and libs need to be added to the final program instead (like before). Some people prefer this. I prefer the convenience library as they are more flexible but they make the compilation a bit longer.

There are additional stuff that could be done, but I think that's enough for one PR. What I can propose then is:

 1. Takes the version number from git. To release, you `git tag`, `./autogen.sh`, `./configure` and `make dist`. It adds a small script that will detect if git is used and if not it will use a version hard-coded in a generated file by `make dist`.

 2. `./configure` is adding libraries to `LIBS`. The result is that everyone is linked with all the libraries. We could add one variable per dependency instead and put them in `Makefile.am`. Sort like:

```
MYSQL_CFLAGS="-I/usr/include/mysql"
MYSQL_LIBS="-lmysql"
AC_SUBST([MYSQL_CFLAGS)]
AC_SUBST([MYSQL_LIBS])
```

And in `Makefile.am`:

```
if USE_MYSQL
libdaemons_la_LIBADD += @MYSQL_LIBS@
libdaemons_la_CFLAGS += @MYSQL_CFLAGS@
endif
```

But I am unsure if it's worth it. Your pick.


vincentbernat:
Forgot to update Travis checks. I have just amended the second commit with that. I also do a "make distcheck". I should move it to the before-last commit. I'll leave it here for now.

vincentbernat:
Hey!

I pushed an additional commit at the end. Again, it is an optional one. It replaces large part of `configure.ac` by the use of `PKG_CHECK_MODULES`. The commit message contains more details, notably with hints on how to override some paths.

This breaks MongoDB support as I am requiring libmongoc-1.0 while it seems the support in pmacct is older. Maybe it would be easy to update it. I didn't look at it much. If everything else in the commit is OK, I will try to make it work.

paololucente:
All commits from this PR have been cherry-picked and merged successfully in master

PR #13, opened 2016-03-04
uacctd.c: switch support from ULOG to NFLOG

vincentbernat:
:ghost: Please, ignore the first commits. I am using #10 as a base since it makes my life easier. Once done, I can rebase the commit to whatever build system we have. Only the last commit is relevant.

NFLOG supports both IPv4 and IPv6. While ULOG is still supported in
recent kernels, NFLOG is supported since 2.6.14 and there is little
point to support both.

Each occurrence of "ulog" has been replaced by "nflog". However,
"uacctd" is still called this way and configuration options remain the
same (but the semantics change a bit for some of them). This should
minimize the disruption for downstreams and users (but users will still
have to update the netfilter rules to use NFLOG instead of ULOG).

This is only lightly tested. I'll do more tests next week. I put it here for early comments.


vincentbernat:
I have asked Travis to whitelist libnetfilter-log-dev

vincentbernat:
I have updated the PR against the current master

PR #86, opened 2016-12-18
* nfprobe: don't shortcut loop on poll() error

vincentbernat:
On exit, nfprobe plugin takes some time to shutdown because we have to
wait the default timeout (60s) before checking if we were asked to
gracefully exit. Instead, don't shortcut the loop on poll() error since
most of the code is correctly guarded a few lines after that.

This seems the simpler solution, but we could also make an exception for
graceful shutdown, or do the graceful shutdown in the signal
handler (like for SQL) or register a dedicated pipe to signal the
graceful shutdown to poll().

### DIFF
--- a/src/nfprobe_plugin/nfprobe_plugin.c
+++ b/src/nfprobe_plugin/nfprobe_plugin.c
@@ -1553,7 +1553,6 @@ sort_version:
   }
 
   for(;;) {
-poll_again:
     status->wakeup = TRUE;
 
     pfd.fd = pipe_fd;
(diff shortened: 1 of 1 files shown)


vincentbernat:
With the current solution, the plugin won't bail out. It still needs the `graceful_shutdown_request` flag to be set, otherwise, the error will be treated as previously (except the Kafka/AMQP stuff will be executed).

Putting everything into a signal handler seems dangereous to me (usually, you have many stuff which are not safe  to do in a signal handler). Let me amend a bit the PR to have solution 2.

PR #87, opened 2016-12-20
Update uacctd.c / interface and direction

Thomas-Hn:
This is the first time I use GitHub and submit something. If this is wrong way, let me know ;)

If you have an router at a geographic site that has many interfaces, it will be hard to analyze traffic in tools like Netflowanalyzer because they are separated by interfaces. We are only interested in what traffic enters the whole site and leave the site through a satellite link. With this, the whole site appear as interface "1" This was the reason for this change. But this is probably not interesting for everyone.

This change sets the ifindex in netflow always to "1"
the direction can be set dynamically via iptables rules. Just add --nflog-prefix <in|out> to the NFLOG target.

### DIFF
--- a/src/uacctd.c
+++ b/src/uacctd.c
@@ -71,13 +71,23 @@ static int nflog_incoming(struct nflog_g_handle *gh, struct nfgenmsg *nfmsg,
   hdr.caplen = MIN(pkt_len, config.snaplen);
   hdr.len = pkt_len;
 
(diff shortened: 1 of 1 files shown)


paololucente:
Hi Thomas, thanks for submitting this. As you also noted, the patch is not of general applicability and hence it can't be merged. But on a positive note it will remain here so that any other person that can benefit from it, can apply it. Thanks for the effort, Paolo

PR #123, opened 2017-03-05
enabling ipv6 support by default (use --disable-ipv6 instead)

fooelisa:


fooelisa:
based on a discussion with paolo, this should be on by default these days

jejenone:
nice!

PR #147, opened 2017-08-11
* add src_vlan and dst_vlan primitives for sFlow

mikejager:
The existing 'vlan' primitive handler for sFlow selects the ingress VLAN ID,
unless it is not present, in which case the egress VLAN ID is used. Two new
primitives ('src_vlan' and 'dst_vlan') are added so that both ingress and
egress VLAN IDs are always available.


mikejager:
https://github.com/pmacct/pmacct/blob/3757ced846926a693ae2e7e8eee27742ec447309/src/pkt_handlers.c#L4661 sets the 'vlan' primitive to the **destination** VLAN ID if the source VLAN ID is zero.

In the case where a frame enters a device untagged on port A, and exits the device tagged with VLAN 1234 on port B, wouldn't what you're suggesting result in the [src_]vlan primitive being incorrectly set to the 1234 rather than 0?

mikejager:
Having `src_vlan` set `COUNT_INT_VLAN` may be confusing for users. They will have requested the `src_vlan` primitive, but this will be replaced with the `vlan` primitive on output.

If you want to avoid creating both a `src_vlan` and `dst_vlan` primitive, it may be better to drop the `src_vlan` primitive entirely; users would request `vlan` and `dst_vlan` instead. What do you think?

paololucente:
Hi Mike,

Re-thinking this after some time: i propose to drop src_ and dst_ prefixes in favour of in_ and out_ ones: it aligns to interface primitives, ie. in_iface and out_iface, and overall it aligns to sFlow. Agree it looks more natural?

Then, second point, backward compatibility (we have been circling around this in the previous comments): we have the existing 'vlan' primitive which we should alias on either of in_vlan or out_vlan. You essentially propose to alias it on in_vlan and i am totally fine with it. I can complement this change with some documentation.

Paolo

mikejager:
Hi @paololucente,

No problem replacing `src_` and `dst_` with `in_` and `out_`.

How to make the change backward compatible is what I'm still unsure about. Assuming we alias `vlan` on to `in_vlan` (which seems to make sense given that the "in" VLAN is the first to be used for `vlan`), are you suggesting replacing `vlan` with `in_vlan` but keeping `vlan` for backwards compatibility (perhaps for sFlow only?).

paololucente:
Hi Mike, perfect about in_ and out_ and entirely correct about the aliasing part.

paololucente (commented):
Closing. Duplicate of #629

PR #151, opened 2017-08-24
Permit use of INET6_ATON() for AF_INET using sql_num_hosts and sql_num_ipv4_inet6aton

gunkaaa:
INET6_ATON() allows representation of both AF_INET (as VARBINARY(4)) and AF_INET6 (as VARBINARY(16)), allowing either to be stored in VARBINARY(16). However, INET6_NTOA() cannot handle the INT(4) representation produced by INET_ATON(). https://dev.mysql.com/doc/refman/5.6/en/miscellaneous-functions.html#function_inet6-aton

This commit allows use of INET6_ATON() instead of INET_ATON() for numerical representation of IPv4 addresses.


gunkaaa:
Hi,

I'm happy to have `sql_num_inetaton`, with an inverse meaning to `sql_num_ipv4_inet6aton`.

In the context of MySQL/MariaDB, there are a few things which need to be done before this can be made default:
* The introduction of several variable-sized fields (the only existing one being `label`) carries heavy performance penalties under both InnoDB (`VARBINARY` is stored off-table) and under MyISAM with `ROW_FORMAT=(DYNAMIC|COMPRESSED)`. `sql/README.mysql` and `sql/pmacct-create-db*.mysql` need to be updated to recommend MyISAM with `ROW_FORMAT=FIXED`, as some quick testing shows performance roughly halving by default; the disk tradeoff should acceptable as `VARBINARY(16)` is only two bytes larger than `CHAR(15)` as used by IPv4 without `sql_num_hosts`.
* #123 enabled IPv6 by default, but there was no change of default host type in `sql/README.(mysql|sqlite3)` or in the accounting table templates from `CHAR(15) NOT NULL` to `CHAR(45) NOT NULL`. With the changes in `sql_mode` to `STRICT_TRANS_TABLES` in MySQL 5.7.5 and MariaDB 10.2.4, this could cause inserts with `sql_multi_values` to fail when IPv6 is accounted for along with IPv4 (previously v6 records were just truncated). 

Given the above, it seems that `sql/README.mysql` should be updated to:
```
* src_host => ip_src (CHAR(45) NOT NULL)
  - or (CHAR(15) NOT NULL, if --disable-ipv6)
  - or (VARBINARY(16) NOT NULL, if sql_num_hosts: true)
  - or (VARBINARY(4) NOT NULL, if sql_num_hosts: true and --disable-ipv6)
  - or (INT(4) UNSIGNED NOT NULL, if sql_num_hosts: true and --disable-ipv6 and sql_num_inetaton: false)
```

Let me know your thoughts and I'll update this PR.

PR #161, opened 2017-10-11
Change parent alive check to work with Docker

hvanderheide:
The origional code checked if the parent PID was equal to 1 (init) to
detect if the parent process died. Docker containers only launch the
required processes which would make the pmacctd core process PID 1,
failing the check.

During initialization of the plugin the core PID is already saved. We
can use this to compare it with the current PID which should have the
same intended effect and still works with Docker.

If this solution is accepted it should be duplicated to the other
plugins.


hvanderheide:
Hi Paolo, thanks for reviewing so quickly. I've implemented the fix on the other plugins. Regards, - Hidde

PR #196, opened 2018-05-03
Fix pcap iterface-map lookup

pldubouilh:
So `eth1` and `eth10` are not mixed up anymore :+1: 
It was mostly an issue when the map was reloaded, as it made libpcap re-registering the same interface.


paololucente:
Hi @pldubouilh , thanks very much for this. May i ask you the favour to to the same PR against the 1.7.1 branch? Otherwise i'll find a way to back port it. Paolo

pldubouilh:
Hey @paololucente - sure doing so now

paololucente:
A bit late but thanks for this @pldubouilh

PR #204, opened 2018-06-06
added support for geoipv2 lat/lon primitives

slash31:
I tried to keep all changes in line with existing structure/format. So far the lat/long lookups seem to work well on a qa/testing cluster of collectors that receive ~200k messages per second at 1:10k sample rate. We're currently using the GeoLite2-City DB file for testing.

The benefit of having lat/lon for us is that our backend datastore (CrateDB) has a geo_point data type, which we can use to directly query against polygons in GeoJASON format.


slash31:
Hi Paolo - no problem at all; do you have a preferred/suggested datatype for a combined lat/lon? I could make it a string in an accepted geo format (GeoJSON or WKT), or an array/object of two doubles.

paololucente:
Hi Aaron, i was only thinking to have one single configuration knob for both latitude and longitude. Then, i am totally with you that we need two 'double' variables - one to store latitude, one to store longitude. Taking as example the MongoDB plugin:

`
if (config.what_to_count_2 & COUNT_SRC_HOST_COORDS) {
  bson_append_double(bson_elem, "lat_ip_src", data->src_ip_lat);
  bson_append_double(bson_elem, "lon_ip_src", data->src_ip_lon);
}

if (config.what_to_count_2 & COUNT_DST_HOST_COORDS) {
  bson_append_double(bson_elem, "lat_ip_dst", data->dst_ip_lat);
  bson_append_double(bson_elem, "lon_ip_dst", data->dst_ip_lon);
}
`

slash31:
Ah, got it. No problem. :)

slash31:
Updated code committed, let me know if this is what you had in mind. I'm obviously new to the codebase, so please don't hold back from pointing out anything that's not correct.

PR #219, opened 2018-08-15
ADD-PATH capability exchange fix for MP-BGP OPEN

FvDxxx:
PMACCTs handled/handles replies to MP-BGP ADD-PATHs incorrectly. It checks only first AFI.SAFI and sets in the reply for last AFI.SAFI RECEIVE(1) if first included SEND(2) or SEND-RECEIVE(3).

Example: Peer with SEND capability for inet unicast (1.1) and inet6 unicast (2.1) sends:
        Optional Parameter: Capability
            Parameter Type: Capability (2)
            Parameter Length: 10
            Capability: Support for Additional Paths
                Type: Support for Additional Paths (69)
                Length: 8
                AFI: IPv4 (1)
                SAFI: Unicast (1)
                Send/Receive: Send (2)
                AFI: IPv6 (2)
                SAFI: Unicast (1)
                Send/Receive: Send (2)
Reply was/is from bgp_msg:
        Optional Parameter: Capability
            Parameter Type: Capability (2)
            Parameter Length: 10
            Capability: Support for Additional Paths
                Type: Support for Additional Paths (69)
                Length: 8
                AFI: IPv4 (1)
                SAFI: Unicast (1)
                Send/Receive: Send (2)
                AFI: IPv6 (2)
                SAFI: Unicast (1)
                Send/Receive: Receive (1)

This patch hopefully fixes it (not too much testing done ...). It checks each AFI.SAFI included in the capability signaled from the peer and only replies with RECEIVE for AFI.SAFI if peer said SEND or SEND-RECEIVE for it.

(Sorry, pasted into http:// from my source ... pull, push ... sigh).

### DIFF
--- a/src/bgp/bgp_msg.c
+++ b/src/bgp/bgp_msg.c
@@ -289,23 +289,37 @@ int bgp_parse_open_msg(struct bgp_msg_data *bmd, char *bgp_packet_ptr, time_t no
 		char *cap_ptr = optcap_ptr+2;
 		struct capability_add_paths cap_data;
 
-		memcpy(&cap_data, cap_ptr, sizeof(cap_data));
-
-		if (online) {
-                  bgp_peer_print(peer, bgp_peer_str, INET6_ADDRSTRLEN);
-		  Log(LOG_INFO, "INFO ( %s/%s ): [%s] Capability: ADD-PATHs [%x] AFI [%x] SAFI [%x] SEND_RECEIVE [%x]\n",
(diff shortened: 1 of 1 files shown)


paololucente:
Hi Markus,

No, no, i totally like the "cap_set approach". And with your comment above i think i have the elements to make the comment (in the code ;-)) a bit more clear - which was all my point about. 

Thank you very much again, look forward seeing you at NLNOG in a couple of weeks.

Paolo

PR #228, opened 2018-09-21
pmacctd: add option 'gre_decap' for IPv4/GRE decapsulation of captured packets

francoijs:
Hi there, 
I propose this option that enables accounting of packets inside a GRE tunnel
---
If set, makes the daemon decapsulate GRE-tunneled traffic.
Packets that are not encapsulated in an IPv4/GRE tunnel are processed as usual.
---


francoijs:
Hi, thanks for the quick feedback!
No problem, I will have a look at the GTP code ASAP

paololucente:
@francoijs , thank you Sir. Paolo

PR #240, opened 2018-10-09
change ip_src/ip_dst field length for IPv6

Rico29:
ipv6 max length is 45.


Rico29:
Hi Paolo. Changes are done.
Regards

PR #256, opened 2019-01-12
Initial JSON UDP Plugin Implementation

whh8b:
This is a simple plugin that will format captured
records as JSON objects and then deliver them
to a listening UDP server.


whh8b:
Hello! pmacct is such a great set of programs and a great tool. We are using it in our research project and had the requirement that we format netflow records as json-objects as part of a data processing pipeline. This little plugin does exactly what we need -- it formats each record as a json object and then sends it to a server over UDP. 

I would love for this to be included if you think that it's useful. I know that there will be things that you want to fix -- please let me know how I can modify the patch so that it meets your high standards.

Thanks again for the great work maintaining and developing such a great suite of tools!

Will

whh8b:
Just FYI: As we emailed about last night, here is an updated pull request for the plugin that includes zeromq support! I hope that this matches your expectations!

As we also talked about in the email, I am waiting on the rebranding to hear your creative thoughts on the new name :-) 

I look forward to working together to get this into shape so that it matches your high-quality standards!

PR #279, opened 2019-04-09
implement multiple outputs to nfprobe_receiver

pldubouilh:
Hey @paololucente !
Here's a PR that allows to set multiple endpoints to `nfprobe_receiver`. Samples are then shipped to a pseudo-randomly selected endpoint. 
The idea is to load balance the processing of samples. Hope it's of interest ! Keep me posted if you think something should be changed.
Cheers


pldubouilh:
thanks for the heads up, I'll investigate nfacctd/tee in addition to pmacct for this then :+1: 
cheers :)

PR #292, opened 2019-06-11
* fix: correct handling of variable length IPFIX fields

nimrody:
nfacctd does not correctly handle variable length IPFIX packets.

The problem stems from the incorrect updating of `remlen` in `resolve_vlen_template()`. Since `len` keeps increasing and holds the current packet length, one cannot update `remlen` by `remlen -= len`. The currently remaining number of bytes in the flow is actually `remlen - len` (assuming `remlen` is not modified like in the suggested patch).

### DIFF
--- a/src/nfv9_template.c
+++ b/src/nfv9_template.c
@@ -1219,7 +1219,7 @@ int resolve_vlen_template(u_char *ptr, u_int16_t remlen, struct template_cache_e
       if (otpl_ptr->tpl_len == IPFIX_VARIABLE_LENGTH) {
 	vlen = TRUE;
 
-	ret = get_ipfix_vlen(ptr+len, remlen, &otpl_ptr->len);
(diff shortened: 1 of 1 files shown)


nimrody:
Hey @paololucente! I suspect there are more bugs in handling variable length IPFIX information elements. How do you test the code? I haven't seen any tests in the repository.

paololucente:
hey @nimrody , thank you and any further code review would be greatly appreciated. And you are right: there are no public tests: i test against both live traffic (general testing, ie. does it crash?) and traces (for sepcific features) but unfortunately neither one nor the other can be made public. Paolo

paololucente:
Hi @nimrody , this is merged. I was wondering whether you could get in touch via unicast email (i don't have your email address) to see what are your other suspicions of things not working well wrt IPFIX & variable length fields. Paolo

PR #302, opened 2019-07-28
misc: fix warnings introduced 0e25ee3

msune:
Move `int no=0` to #ifdef section, so that both Linux and non-Linux
compilations work without warnings.


msune:
I don't have a non Linux device here, so please double check BSD* is not broken again.

PR #304, opened 2019-07-30
Fix more warnings detected by GCC8 and clang

msune:
Fix more warnings detectedby gcc8 and clang.

Please review thoroughly.

PR #306, opened 2019-08-03
Major refactor: add header guards, remove {EXT, _EXPORT, _C}, fix GCC/clang warnings

msune:
This PR is a long one and superseed #304. The patchset does:

* Fixes configure args in .travis CI, so that all plugins are enabled in "all-plugins" env
* Fixes a log of GCC8, and some clang warnings
* **Removes all `#define EXT #undef`, `_EXPORT`** etc... This is a major change. To do that:
   * All headers now have standard guards `ifndef X #define X`
   * Most global variables declarations have been moved to .c files, and left fwd declarations as exter in headers
   * Fix conditional compilation of some plugins

This PR needs *extensive review*


msune:
Rebased.

**Added `-Wall -Werror` for non-debug builds** and fixed all remaining warnings in gcc4.9.

PR #316, opened 2019-08-27
FIX SQL schemas identation

rouba002:
This is a little messy history of this commit, but still I hope this will help.
Unified identation of SQL schema files to 4 spaces. Previously mix of TAB and SPACE.


rouba002:
Hi @paololucente 
1) The identation is incosistent, tabs, 4 spaces, sometimes just 2. So I tried to unify them., I am ok with any suggested option
2) Yes you are right, as I am not as skilled I tried to remove that particular commit from history, but obviously failed
If I fail to do that I will create a new fork and fix that identation on fresh.

rouba

paololucente:
Hi @rouba002 , ack about point 1. Please start a new commit to only handle the indentation. Thanks, Paolo

PR #331, opened 2019-10-20
Replace ABSTIME with to_timestamp in PG sqls

### LINKED ISSUE #329: pgsql plugin incompatible with pg12
Came across this when working on some new configs:
```
server_1  | INFO ( all/pgsql ): *** Purging cache - START (PID: 183) ***
server_1  | ERROR ( all/pgsql ): ERROR:  function abstime(integer) does not exist
server_1  | LINE 1: ...to, tag, class_id, packets, bytes, flows) VALUES (ABSTIME(15...
server_1  |                                                              ^
server_1  | HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
server_1  |
server_1  |
```
The PG12 release notes states [1]:
> * Remove data types abstime, reltime, and tinterval (Andres Freund)
>   These are obsoleted by SQL-standard types such as timestamp.

Happy to try out a PR if you have a way you'd like to see this implemented.

[1]: https://www.postgresql.org/docs/12/release-12.html

mmlb:
ABSTIME was dropped in PG commit da6a8d01d391eab45c4b3e0043a1b2b31072f5f
and released in PG12. ABSTIME has been marked deprecated for a *long*
time, to_timestamp is the supported equivalent.

to_timestamp goes back to at least 8.4 which is the PG version shipped
in RHEL/CentOS 6 and the minimum version supported via configure.ac so
all is good.

Closes #329

### DIFF
--- a/src/pgsql_plugin.c
+++ b/src/pgsql_plugin.c
@@ -631,7 +631,7 @@ int PG_evaluate_history(int primitive)
       strncat(where[primitive].string, " AND ", sizeof(where[primitive].string));
     }
     if (!config.timestamps_since_epoch)
-      strncat(where[primitive].string, "ABSTIME(%u)::Timestamp::Timestamp without time zone = ", SPACELEFT(where[primitive].string));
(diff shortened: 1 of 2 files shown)


mmlb:
no problem :+1:

PR #335, opened 2019-10-29
Missing log attribute

rbarazzutti:
Check the existence of peer->log before accessing attributes in order to avoid to have a segfault. Emit a log warning message when it occurs.


rbarazzutti:
Thanks!

PR #337, opened 2019-11-04
add config option: print_allow_empty_file = true|false

setup74:
to create empty file when no cache entry to purge out

My system uses old pmacct's behavior -- to generate empty file even not data to purge exists -- 
to chech the proof-of-healthyness of sfacctd by the existence of data file (configured per-minute).
But,  new sfacctd omits data file without data. So option is added to behave as old...

### DIFF
--- a/src/cfg.c
+++ b/src/cfg.c
@@ -130,6 +130,7 @@ static const struct _dictionary_line dictionary[] = {
   {"print_output_separator", cfg_key_print_output_separator},
   {"print_output_custom_lib", cfg_key_print_output_custom_lib},
   {"print_output_custom_cfg_file", cfg_key_print_output_custom_cfg_file},
+  {"print_allow_empty_file", cfg_key_print_allow_empty_file},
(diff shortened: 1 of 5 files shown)


setup74:
Dear @paololucente,

1)  It sounds good. I've changed to to "print_write_empty_file"

2a) I rechecked and index > 0, pqq_ptr > 0 is needless, so rollbacked.
      (just from my old experiences on memcpy()'s with size=0,
        but i've tested and checked memcpy() with size=0 is ok)

2b) the save_index > 0 test is protect the use of queue[0] at next lines,
       on the empty casse the use of queue[0]->* caused segfault.
       But i've not confirmed that queue[0] should be NULL or could be some garbage pointer,
       so I just tested save_index > 0 instead of queue[0].
       If queue[0] is always NULL on empty case, the test logic would be good to be queue[0] != NULL

I'v applied 1 and 2a and commited to github.

Thanks.

paololucente:
Wonderful, thank you @setup74 . Merging. Paolo

PR #344, opened 2019-11-20
add non blocking i/o

jaredmauch:
set non-blocking I/O on socket


jaredmauch:
I do not expect any issues from this patch, but hope to provide a report from deployment of the code in our environment later this week

Issue #348, opened 2019-12-02
bmp: use SO_KEEPALIVE


jaredmauch:
Sockets may stick around after the router goes down or disconnects.  set SO_KEEPALIVE to allow them to be garbage collected if the TCP session goes away

jaredmauch:
https://github.com/pmacct/pmacct/pull/349

paololucente:
Merged.

PR #349, opened 2019-12-02
#348 pmacct/pmacct SO_KEEPALIVE

jaredmauch:


jaredmauch:
We have tested this in production and it seems to resolve what we were seeing.

PR #397, opened 2020-05-24
Added new attributes and BMP message type 5 and 6

graf3net:
### Checklist

I have:


graf3net:
Hi Paolo, as requested, I updated the document not only with the added BGP ADD-Path, AIGP, Prefix-SID attributes but also with the updated from the BMP hackathon such as Peer RD, path marking and message type 5 and 6.

PR #399, opened 2020-06-02
ci: use dedicated jobs per {docker, config_flags}

msune:
### Short description

Prior to this commit there were only 2 jobs being launched, in which
all possible combinations of pmacct flags were compiled.

This had two fundamental problems:

* Jobs could not run in parallel so could not exploit parallel jobs
  of travis CI (even in free tier)
* If a combination was failing, the entire job was failing

In addition all the log was in the same _very long_ job, making it
difficult to debug.

Unfortunately, parallel jobs is not improving pipeline times, since
docker generation - which takes > 70% of time - cannot be reused
across jobs.

This should be addressed in a later commit, in which two different stages
are defined.

This commit builds on top of:

* Moves dependencies to Dockerfile
* Moves all related files from docker/ to ci/. This is to avoid users
  thinking that docker file is a pmacct docker => btw this might be
  a good idea; someone thought of adding a pmacct public layer?
* Refactors CI to use a job per tuple {docker, config_flags}. It has
  a single stage and it prefixes all jobs with name:
   - Ubuntu Bionic
   - Centos 8.1
* Tries to simplify the CI code
* Adds minor trick to pmacct ./configure to show config.log in case
  of failure

### Checklist

I have:


msune:
Several comments:

Take a look at the job structure, and let me know your thoughts: https://travis-ci.org/github/pmacct/pmacct/builds/693668890 

* Bonus: negative diff.
* Multi-stage, with docker caching is here: https://github.com/msune/pmacct/commit/76952800330ab386e2277fe7636664ab9a53be43. But it seems an external "storage" is needed (and btw, care needs to not break pipelines in PRs from other repos)... this is annoying.
* Did you consider having a Dockerfile with say, Debian + pmacct "ready to go"?

msune:
Added `wget` retrial backoff. Pipelines sometimes were failing retrieves packages using `wget` when preparing the container.

PR #404, opened 2020-06-07
Docker improvements: reduced container size, {bleeding edge, latest, vX.Y.Z}, pmacctd container

msune:
### Short description

This patchset:

* Reduces the `base` container size roughly 70%
* Uploads `base` docker image with three different tags:
  - `latest`: always the latest *stable release*. It is updated on every new release
  - `vX.Y.Z`: always presenet, once published. Fixed tag for the release vX.Y.Z
  - `bleeding-edge`: only for the brave ones. Snapshot of the current master HEAD
* Creates `pmacctd` container based on `base`. This container, in contrast to `base` has the daemon as an entry point and expects `/etc/pmacct/pmacctd.conf` to exist (being mounted as a volume). Same tags as `base` are pushed.

After the integration of this PR, `bleeding-edge` containers will start to be generated.

### Checklist
I have:


msune:
For daemons that listen, either:

* `EXPOSE` directive in Dockerfile shall be added (and list all possible ports - this is particularly important for `base`)
* docker needs to run with host network.

PR #417, opened 2020-07-11
Debug messages related to Netflow parsing

vphatarp:
Because of BGP policy changes I was seeing issues on the network which required me to filter out which Flow Destination IP addresses were not showing up any match on BGP tables learned by the BGP daemon. Earlier I used to debug this issue by dumping entire BGP tables on Nfacctd Collector and generating the flow matrix file with Ip Src/Dst listed and then doing a correlation between the 2. 
I believe some debug commands related to Netflow parsing would ease debugging efforts needed and provide easier drill downs. It would also help to push the logs to ELK or similar Log Analyzers in order to debug the issue faster. I have added debug commands to the code which lists the following details:
1) IP Address of the device that sourced the Netflow packet.
2) BGP PeerIP which would be used to perform the lookups.(Helps in scenarios where RRs are used)
3)  BgpNextHop for the flow as seen in the Netflow packet.
4) Source IP address of the Flow.
5) Destination IP address of the Flow.
6) Whether a lookup Match was found or not.

I have added a new cfg command for a conditional trigger of the debug commands. Only when debug and debug_netflow_parsing are both set to true, messages will be pushed to the log files. Currently debug messages have been added only to IPv4 and IPv6 flows.

Please let me know if any changes are needed from my end.
Regards,
vphatarp 

### Checklist

I have:

### DIFF
--- a/src/bgp/bgp_lookup.c
+++ b/src/bgp/bgp_lookup.c
@@ -1,4 +1,4 @@
-/*
+/*  
     pmacct (Promiscuous mode IP Accounting package)
     pmacct is Copyright (C) 2003-2020 by Paolo Lucente
 */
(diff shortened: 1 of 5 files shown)


patrickjahns:
@paololucente 
Any update on the process of merging the PR ?

PR #420, opened 2020-07-18
Autogenerate pmacct-build.h

msune:
Maintaining, by hand, `pmacct-build.h` is a waste of Paolo's
precious time. This commit autogenerates - in the same format - the `pmacct-build.h` build string.

It's highly unlikely this change will make it in. It has two inherent
problems, pretty much unavoidable (if someone has ideas, please share):

* If .git/ is not present, the string cannot be deduced - obviously
* If `./configure` is not repeated when new commits are pulled,
  the build of the last `./configure` would appear. Even more, if
  old objects are present, they will not be recompiled.

Anyhow, at least Paolo can extract the shell commands to autogen
the file :D

### Checklist
I have:


msune:
wrt to your comments, first I have to say the PR is half a joke and half serious. I just wanted to give you the commands. 

Second, I realised the commit message is indeed misleading. The initial intention was to autogen `pmacct-build.h` from an `pmacct-build.h.in`, but then I went the easy path of `AC_DEFINE` as propsective.

To your questions:

>Guess a commit involves a certain file and someone does a 'git pull' after that; the configure script is not guaranteed to be called. Actually there is a risk that the build remains stuck to the first time you did a 'git clone'. Agree? Disagree? If agree, would you see a way around it? I think this is the key point.

Yes, sort of (autoreconf would exec if some `Makefile.am` would change) but the problem for compiled objs would stand still:

>>If ./configure is not repeated when new commits are pulled, the build of the last ./configure would appear. Even more, if old objects are present, they will not be recompiled.

On the serious side of things...

Why do you need a `PMACCT_BUILD` at all? I understand the need for this variable before `git`. After `git`, the commit `hash` is unequivocally identifying the precise build/commit. Most projects use that as a build number, and that has the advantage that `git log <hash>` gives immediately the log of the commit, without a translation in between.

For releases, this is actually not a problem - IMHO - as the git tag has a 1:1 relation to the commit, hence again, you can `git log <tag>`. So that would solve the problem of the tarballs for releases, as if .`git/` is not find, the BUILD string could be `RELEASE` or s/t along those lines.

Regardless of what actually identifies the build, current format or git hash, the problem of already cloned (and built) repos, would be there.

I am pretty sure there must be a way to cache locally the commit ,and in a Makfefile.am, if a new commit now exists, force an autoreconf or alike... For old compiled objects, then `pmacct-build.h.in` **shall** be used to solve it.

If you want me to truly look into it, just let me know. You can also close the PR if you don't see the need to bother... Really, no strings attached :D

paololucente:
Hey Marc,

> wrt to your comments, first I have to say the PR is half a joke and half serious. I just wanted to give you the commands.

Yes, yes i got that ❤️ 
 
> Why do you need a `PMACCT_BUILD` at all? I understand the need for this variable before `git`. After `git`, the commit `hash` is unequivocally identifying the precise build/commit. Most projects use that as a build number, and that has the advantage that `git log <hash>` gives immediately the log of the commit, without a translation in between.
> 
> For releases, this is actually not a problem - IMHO - as the git tag has a 1:1 relation to the commit, hence again, you can `git log <tag>`. So that would solve the problem of the tarballs for releases, as if .`git/` is not find, the BUILD string could be `RELEASE` or s/t along those lines.

The idea was to make it super simple for people, ie. when opening an issue or reporting a bug a 'nfacctd -V' would capture all needed info to start a meaningful troubleshooting / debugging (version, commit, compiling options, compiler info and system info). Reason being: not everybody cloning from GitHub can do much beyond 'git clone'. This is why referencing the hash (or the beginning of the hash better) instead of using the current jargon would work too.

> If you want me to truly look into it, just let me know. You can also close the PR if you don't see the need to bother... Really, no strings attached :D

I really was looking at this in the past myself but stumbled pretty much into what you did stumble. I would say this is a nice to have thing (as opposed to a necessary one) and i always said to myself there is better use of time. But you have voodoo powers, i know; so i would say: if it looks like something that can be solved quickly, great. Otherwise if research and thinking are needed then perhaps, yeah, we can defer to the next pandemic boredom.

Paolo

msune:
Solved,

Take a look at the commit message itself:

https://github.com/pmacct/pmacct/commit/b99d2428eeaee61297d7593e8a78452f72fc6430

They key is:

https://github.com/pmacct/pmacct/commit/b99d2428eeaee61297d7593e8a78452f72fc6430#diff-480477e89f9b6ddafb30c4383dcdd705R20

Regards

msune:
Added the git hash (short form):

```
marc@Mriya:~/personal/pmacct$ src/nfacctd -v
src/nfacctd: invalid option -- 'v'
NetFlow Accounting Daemon, nfacctd 1.7.6-git (20200725-1 [a10602f0])
```

PR #429, opened 2020-09-30
BMP leak fix

rbarazzutti:
### Short description

After some investigations, I was able to spot some leaks that were causing pmbmpd memory usage to consume a bit too much memory.
### Checklist

I have:


paololucente (commented):
Hi Raphael ( @rbarazzutti ), thank you very much for this contribution. All looks good & will merge imminently. Paolo

PR #434, opened 2020-10-07
adding crc32 as another method of hashing for tee plugin

edge-intelligence:
I have added crc32 hash algorithm as an option for tee_plugin.
It's aim is to get a better balancing to receiver nodes.

example of usage receivers.lst:
id=1    ip=127.0.0.1:2001,127.0.0.1:2002,127.0.0.1:2003 balance-alg=hash-crc32

### DIFF
--- a/examples/tee_receivers.lst.example
+++ b/examples/tee_receivers.lst.example
@@ -22,7 +22,8 @@
 !			robin, 'hash-tag' hashing of tag (pre_tag_map)
 !			against the number of receivers in pool, 'hash-agent'
 !			hashing of the exporter/agent IP address against
-!			the number of receivers in pool.
+!			the number of receivers in pool,
+!           'hash-crc32' same as hash-agent but uses crc32 as hash function.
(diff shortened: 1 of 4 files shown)


paololucente (commented):
Hi @edge-intelligence , thank you Guys, much appreciated. Surely an interesting feature. Going to merge imminently. Paolo

paololucente:
Hi @edge-intelligence ,

I did work out the IPv6 case and for the task i adapted an existing addr_hash() function to work against sockaddr structures, sa_hash(). See here the commit: https://github.com/pmacct/pmacct/commit/78970c571d54a7a1dc512edfd5faac0c66b5116a 

In theory this can be used seamlessly for IPv4 too but i did not want to touch that part without some testing first: if you think this is useful and have a way to test its robustness, please let me know your thoughts and maybe open a new PR with the updated code? Should find you positive probably at that point the feature should be renamed to something else than crc32, dunno, 'hash_exporter_ip' or something like that.

Paolo

PR #440, opened 2020-11-04
Add git submodule dependency builds. Add -lcdada as dependency 

msune:
### Short description

Dependencies are always a bit of PITA. This patchset extends the build system to support building external libraries and statically link them to `pmacct` for the ease of users.

The fundamental design principle has been to _keep the "user interface"_ exactly the same:

```
git clone
cd pmacct
./autogen.sh
./configure
sudo make install
```

As per discussed with @paololucente, this commit:

* Introduces in `src/external_builds` directory. Under this directory:
  *  git submodules can be added for each dependency
  * `Makefile.am` contains a single target `submodule_deps` that list the submodule recipes (`make` targets) to build. Make sure to add also cleanup functions.
  * Dependencies are only compiled when needed, including when the submodule commit changes (even locally without committing) 
* Adds `libcdada` as a dependency of `pmacct` (all) (currently pointing to v0.3.2) and a show-case of the submodule concept. This also adds the dependency of `libstdc++` (dynamic)

##### Use-cases

1. Regular users using `git`: build system works exactly the same as before.
2. Tarballs: build system behaves the same as before. Users need to compile `libcdada` and install it in the system
3. OS package maintainers: can use 1) or opt to use `--without-external-deps` to skip completely building the submodules.

For 2, 3, `-lcdada` is checked as a dependency in `configure.ac`. Intree and OutTree compilation is tested (CI) and checked. `make clean` and `make maintainer clean` manually checked in both cases.

##### Other comments

* I would still move `src/external_libs/` -> `external_libs/`. To me it looks cleaner, and will still work.
* I would suggest to explore the same approach for some of the _hard to compile_ dependencies of `pmacct`, those not available in the OS pkg managers. That might save some time to users.
* The `Makefile.am` should be self-explanatory on how to add more deps (just add more targets), but let me know if you want extra documentation.

### Checklist

I have:


msune:
Btw, I would suggest to create a mirror of libcdada and point `.gitmodules` there, to have it under your control

paololucente (commented):
Thank you Marc ( @msune ) for your nice library project and this amazing contribution. As we discussed this will help monumentally in streamlining the use of certain basic structures, ie. hashes, lists, etc. in pmacct. I will do a bit of testing at my end and merge asap. Paolo

PR #462, opened 2021-01-15
Fix build for *BSD systems, gcc-4.2 and submodule updates

msune:
### Short description

This MR addresses some portability issues found while compiling in openBSD, with a rather old gcc-4.2. It fixes:

* Substitutes `$(shell <cmd>)` by `` in all Makefiles, fixing `pmacct-build.h` generation and submodule compilation. 
* Bumps `libcdada` dependency to v0.3.4, which fixes compilation warnings in ancient gcc-4.2
* Fixes compilation in *BSD due to missing `pthread.h` in `bgp.h`
* Fixes bug in which, when getting new pmacct commits (`git pull`) on an already cloned repository, and when those commits where updating submodules, the deduction on whether the submodule deps had to be compiled or not was incorrect.

This MR has been tested in:

* Linux (debian9)
* OpenBSD 6.8

### Checklist

I have:

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -31,7 +31,8 @@ pmacct-build.h_h:
 			echo "`git rev-parse HEAD`" > .pmacct-build.h; \
 			BUILD_DATE="`git log -1 --date=short --pretty=format:%ad | tr -d "-"`"; \
 			BUILD_AUX="`git log -1 --date=short --pretty=format:%ad`";\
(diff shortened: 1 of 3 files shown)


msune:
MInd (developers only), that:

>To prevent `make` to sync the commit - only for developers define:
>
>```
>export PMACCT_EXT_LIBS_DONT_SYNC=1
>```

paololucente (commented):
Thank you Marc ( @msune ) for another bombing PR! 🥇

PR #472, opened 2021-02-23
Update ndpi_util.c

SeeMyPing:
Correction : ndpi_finalize_initialization name.

### Short description
Correction of ndpi_finalize_initialization call line 82.

### Checklist

I have:


paololucente (commented):
Hi Rostan ( @SeeMyPing ),

I am myself compiling against nDPI 3.2 stable and - although i see the misspelling there - that function, spelled that way, is in the library. What version of nDPI are you working against? I sense that probably some more checking against NDPI_MAJOR and NDPI_MINOR may be needed if the spelling has been corrected in some more recent stable release.

Paolo

PR #481, opened 2021-03-12
Update QUICKSTART.md

yuriskinfo:
Fixed links according to the  Markdown syntax (link itself goes in () after the link's name in []).

### Short description
As said above - broken links fixing.

### Checklist

I have:

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -1328,8 +1328,8 @@ Some notes:
   global,sasl.password,Endpoint=sb://.XXX.windows.net/;SharedAccessKeyName=YYY;SharedAccessKey=ZZZ;EntityPath=JJJ
   ```
 
-1. When integrating pmacct with the PNDA framework ( [http://www.pnda.io/]() ), a guide on how to make the two communicate has been written-up and published at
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Hi Yuri ( @yuriskinfo ),

Thanks very much for this contribution and for opening issue #482 . I will most probably not merge this one as i am on the verge to roll back from QUICKSTART.md to QUICKSTART text document.

QUICKSTART.md was contributed a couple of weeks ago and i merged it few days ago; i am quickly realising that the Markdown presentation on GitHub is not optimal for the purpose of the document. Main issue being that i can't anymore point people to specific sections of the document, ie. https://github.com/pmacct/pmacct/blob/master/CONFIG-KEYS#L36-#L38 , with the interpreted Markdown nor with the raw file view. Should you have got any ideas on this, we may try to save the Markdown layout.

Paolo

yuriskinfo:
I see, yes, in that case fixing something later to be removed sounds pointless, so I go ahead and close this.
You're correct, Github Markdown flavor does not allow linking inside the page except to the headers, which is not what you are looking for. So, yes, using line numbers inside text file will be better suited.

PR #489, opened 2021-04-02
Add configurable start offset for 'decode_arista_trailer' option

floatingstatic:
### Short description
Arista has changed the structure of the trailer format in recent releases of EOS by appending additional fields such that the byte offset used to decode the output interface from the Arista trailer (#201) is now offset by 27 bytes (EOS 4.25) instead of 8 bytes (EOS 4.21 and earlier) from the end of the packet. To make this backwards compatible we should add a configurable byte offset to mark the start of the trailer from the end of a packet with a default value of 8 to maintain backwards compatibility. Future releases of EOS will likely append additional fields and `arista_trailer_offset` will allow one to compensate for any changes to the trailer structure.

### Checklist
I have:

### DIFF
--- a/CONFIG-KEYS
+++ b/CONFIG-KEYS
@@ -1883,6 +1883,10 @@ VALUES:         [ true | false ]
 DESC:           Decode output interfaces data present on trailer of libpcap packets passed by Arista.
 DEFAULT:        false
 
+KEY:            arista_trailer_offset [GLOBAL, PMACCTD_ONLY]
+DESC:           When 'decode_arista_trailer' is enabled, set the byte offset from the end of the packet to indicate where the trailer starts.
+DEFAULT:        8
+
 KEY:		[ nfacctd_mcast_groups | sfacctd_mcast_groups ] [GLOBAL, NO_PMACCTD, NO_UACCTD]
 DESC:		Defines one or more IPv4/IPv6 multicast groups to be joined by the daemon. If more groups are
 		supplied, they are expected comma separated. A maximum of 20 multicast groups may be joined by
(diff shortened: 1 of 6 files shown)


floatingstatic:
> I think we should add a check for `config.arista_trailer_offset` as well here, such as:
> `if (config.decode_arista_trailer && config.arista_trailer_offset) {`

Because we have a default value I think we can omit this, but I do not think it would hurt to add the extra check. Let me add this.

paololucente (commented):
Hi Jeremiah ( @floatingstatic ),

Thanks very much for this contribution. I will merge soon but also have one comment for your review: having two knobs, one true / false (decode_arista_trailer), one to express a non-zero offset value (arista_trailer_offset), is a waste. We could have just one, ie. arista_trailer_offset, with the logic you implemented and retire the other / old one. I was thinking to backward compatibility but in any case this check `if (config.decode_arista_trailer && config.arista_trailer_offset) {` kills it in any case. Any objections?

Paolo

floatingstatic:
Hi @paololucente , thank you for the feedback. I'm glad you raised this. I'm honestly wondering if anyone outside of my employer (Fastly) is actually using `decode_arista_trailer`. Given the `decode_arista_trailer` is broken for recent EOS releases and no one has complained to you I suspect we are the only ones 😄 

What I would propose, if you are in agreement, is that we remove `decode_arista_trailer` (boolean) completely from the source. In its place we use the new `arista_trailer_offset` (unsigned integer) config option I am adding here, which by default will be set to `0` instead of `8`. We will use a conditional to match `config.arista_trailer_offset` being a non-zero value to get into the relevant block of code that decodes the trailer to get the output interface as I have implemented in the PR. Does that work for you?

paololucente:
Yes, that does work perfectly, we are aligned - you did also fill some blanks that i did not want to fill publicly :-D Last bit, do you want to take up the commit or shall i? What does work for you, works for me.

floatingstatic:
Cool, looks like you made some edits. LGTM. Thanks again!

PR #491, opened 2021-04-08
bmp: fix using unconfirmed BGP ADD-PATH capability

shaitan:
### Short description

`pmbmpd` runs bgp in "offline" mode. Therefore for OPEN message it doesn't confirm ADD-PATH capabilities to peers. As a result bgp peers don't send path identifiers in UPDATE messages.

According to [RFC 7911](https://tools.ietf.org/html/rfc7911) this confirmation is required:

> A BGP speaker MUST follow the procedures defined in [RFC4271] when
   generating an UPDATE message for a particular <AFI, SAFI> to a peer
   unless the BGP speaker advertises the ADD-PATH Capability to the peer
   indicating its ability to send multiple paths for the <AFI, SAFI>,
   and also receives the ADD-PATH Capability from the peer indicating
   its ability to receive multiple paths for the <AFI, SAFI>, in which
   case the speaker MUST generate a route update for the <AFI, SAFI>
   based on the combination of the address prefix and the Path
   Identifier, and use the extended NLRI encodings specified in this
   document.  The peer SHALL act accordingly in processing an UPDATE
   message related to a particular <AFI, SAFI>.

When `pmbmpd` parses UPDATE messages, it incorrectly tries to get path identifiers from them.

This PR disables this capability for "offline" mode.

### Checklist

I have:

### DIFF
--- a/src/bgp/bgp_msg.c
+++ b/src/bgp/bgp_msg.c
@@ -305,7 +305,7 @@ int bgp_parse_open_msg(struct bgp_msg_data *bmd, char *bgp_packet_ptr, time_t no
 			cap_data.sndrcv);
 		  }
 		  if ((cap_data.sndrcv == 2 /* send */) || (cap_data.sndrcv == 3 /* send and receive */)) {
-		    peer->cap_add_paths[ntohs(cap_data.afi)][cap_data.safi] = TRUE; 
+		    peer->cap_add_paths[ntohs(cap_data.afi)][cap_data.safi] = online;
 		    if (online) {
 		      if (!cap_set) {
 			/* we need to send BGP_CAPABILITY_ADD_PATHS first */


paololucente (commented):
Hi Kirill ( @shaitan ),

May i ask you a bit more context around this patch? What the code does right now is to mark true ADD-PATH if it is part of the BGP Open carried by the BMP Peer Up message; then, since BMP is uni-directional by nature, (or `offline`) a reply packet is not crafted. Are you saying that there are implications in how subsequent BGP Update packets carried by BMP Route Monitoring are parsed?

Paolo

shaitan:
Hi Paolo!

> Are you saying that there are implications in how subsequent BGP Update packets carried by BMP Route Monitoring are parsed?

Exactly, these implications are made by [the code](https://github.com/pmacct/pmacct/blob/56281c19fa4058d87ae36673677c16ab2745771f/src/bgp/bgp_msg.c#L1081-L1086).

paololucente:
Hi Kirill,

I see, indeed: the capability exchange influences how BGP Update messages are parsed, clear. So ADD-PATH is received in the BGP Open, feature is enabled when parsing BGP Update messages. What i am struggling with is the original description of the problem, i can't make sense of it `Therefore for OPEN message it doesn't confirm ADD-PATH capabilities to peers. As a result bgp peers don't send path identifiers in UPDATE messages.`.

Let us take a different approach. See here: https://github.com/pmacct/pmacct/blob/master/QUICKSTART#L3034-#L3044 . Can you please produce me a brief BMP trace with the bmp_play tool so that i can reproduce the current issue in the code? I am unfortunately still unable to see it.

Paolo

shaitan:
Sorry for the delay with answer.

> What i am struggling with is the original description of the problem, i can't make sense of it Therefore for OPEN message it doesn't confirm ADD-PATH capabilities to peers. As a result bgp peers don't send path identifiers in UPDATE messages..

According to [RFC 7911](https://tools.ietf.org/html/rfc7911), this capability should be confirmed by the peer:
> and also receives the ADD-PATH Capability from the peer indicating
its ability to receive multiple paths for the <AFI, SAFI>

Since BMP doesn't send a reply packet, the capability is considered as unconfirmed by BGP speaker and it doesn't send path identifier with Update messages.

More details in screenshots:

Open message that contains the ADD-PATH Capability:
Screen Shot 2021-05-18 at 12 46 22

Update message received from this BGP speaker that doesn't include path identifier (only prefixes)
Screen Shot 2021-05-18 at 12 37 26

And also raw bgp packages: [bmp.zip](https://github.com/pmacct/pmacct/files/6500288/bmp.zip)

I tried to use bmp_play.py, but it creates big dump with all messages from all speakers.

paololucente:
Hi Kirill,

Let's focus on the BMP Peer Up message, your first screenshot. If you see in there, there are two BGP OPEN messages: one captures the BGP OPEN that the remote peer sends to the BMP exporter, the other captures the BGP OPEN that the BMP exporter sends back to the remote peer. One of the two OPENs does include the ADD-PATH capability, i bet if you go explore the other one, it will not have it. But, you see, this pertains to the capability supported by the two peers and the way they are configured, in other words BMP has nothing to do with the negotiation and the subsequent BGP UPDATEs, BMP does only report what (roughly) happened in BGP.

Paolo

shaitan:
Hi Paolo,

You are right! The only BGP OPEN does include the ADD-PATH capability and the other one doesn't include it. Thank you for bringing to light that BMP do nothing with the negotiation, now I see the point.

There is still a problem with parsing such BGP UPDATEs. Maybe you have a more elegant solution for it? I can suggest that BMP could analyze the negotiation and enables the ADD-PATH capability based on it.

paololucente:
Hi Kirill,

Fantastic, we are in sync on the role of BMP with regards to ADD-PATH negotiation. With regards to parsing the BGP Update, can you confirm me what version of pmacct you are running and tell me exactly what issue you are running into?

This is the logic implemented: when the Peer Up message is received, pmacct looks into both BGP Open messages and does a sort of reconciliation of the two a-la "do i see a capability in both BGP Open messages? Then it must be true; if i did see it only in one of them then it may be not true so let's disable the capability". Now, while reviewing the code for this issue i bumped in this: https://github.com/pmacct/pmacct/blob/master/src/bmp/bmp_util.c#L207-#L210 , essentially this check is disabled for ADD-PATH and i am sure there must be some further background than what is described (ie. buggy BMP exporter implementation): can you please de-comment line 210, essentially re-enable the check, re-compile and see if that does for you? If so, we'll take it from there; otherwise please explain me better which situation you are running into.

Paolo

PR #492, opened 2021-04-08
Add configurable pcap_arista_trailer_flag_value option

floatingstatic:
### Short description
In applying #489 to a broader set of hardware in my network I observed that certain chipsets set additional flags in the Arista trailer. `1` which is hardcoded in the source as the expected value to tell pmacct "the output interface field is present in the trailer" ends up not being correct the correct value and needs tuning along with the trailer start offset. For example T2 chipsets would utilize `1` but this should actually be `7` on Jericho. Given it is uncertain what this value should be and it may change over time we should make this configurable as well.

### Checklist
I have:
- [x ] included documentation (including possible behaviour changes)

### DIFF
--- a/CONFIG-KEYS
+++ b/CONFIG-KEYS
@@ -202,6 +202,11 @@ DESC:		Arista does set a trailer structure to convey extra info (ie. output inte
 		where the trailer starts.
 DEFAULT:	none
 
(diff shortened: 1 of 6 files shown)


paololucente (commented):
Wonderful @floatingstatic , i believe this patch dogmatically. Will merge it shortly. Paolo

Issue #493, opened 2021-04-09
tee_source_ip no longer works as of ver >= 1.7.6


floatingstatic:
**Description**
It appears that as of [this commit](https://github.com/pmacct/pmacct/commit/a35e3c4d4f5b4f96cc008a4ad8004f3ed04fab11) the `tee_source_ip` configuration no longer works. I am using the tee plugin with nfacctd. My nfacctd.conf is very simple (ip addresses sanitized from their actual values):

```
plugins: tee
tee_receivers: /persist/sys/pmacct/receivers
tee_source_ip: 203.0.113.0
```

`/persist/sys/pmacct/receivers` looks something like this:

```
!
! tee_receivers: Tee receivers map
!

! ips needs to be stringified in the form `ip:port,ip:port,` (dangling coma allowed)
! balance-alg=rr will load balance betweeen the destinations. no effect with 1 ip

!
id=1	ip=203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,203.0.113.254:4739,  balance-alg=rr
```

Note in the above the same receiver IP is used multiple times just to generate hash entropy for my load balancing application (not really relevant to the problem).

I observe this in nfacctd running in debug mode:

```
DEBUG ( default/core ): Received NetFlow/IPFIX packet from [203.0.113.0:58849] version [10] seqno [2185176]
DEBUG ( default/core ): Received NetFlow/IPFIX packet from [203.0.113.0:58849] version [10] seqno [2185182]
DEBUG ( default_tee/tee ): Sending NetFlow/IPFIX packet from [203.0.113.0:58849] seqno [2185158] to [203.0.113.254:4739]
DEBUG ( default_tee/tee ): Sending NetFlow/IPFIX packet from [203.0.113.0:58849] seqno [2185164] to [203.0.113.254:4739]
DEBUG ( default_tee/tee ): Sending NetFlow/IPFIX packet from [203.0.113.0:58849] seqno [2185170] to [203.0.113.254:4739]
DEBUG ( default_tee/tee ): Sending NetFlow/IPFIX packet from [203.0.113.0:58849] seqno [2185176] to [203.0.113.254:4739]
[...]

PR #495, opened 2021-04-11
Fix tee_source_ip - issue #493

floatingstatic:
### Short description
Fixes issue #493 by partially reverting https://github.com/pmacct/pmacct/commit/a35e3c4d4f5b4f96cc008a4ad8004f3ed04fab11

Before this patch using `tee_source_ip: 203.0.113.2`:

```
12:27:56.503088 Out 74:83:ef:1e:7f:99 ethertype IPv4 (0x0800), length 516: 172.24.12.3.38956 > 203.0.113.254.4739: UDP, length 472
12:27:56.503106 Out 74:83:ef:1e:7f:99 ethertype IPv4 (0x0800), length 516: 172.24.12.3.41767 > 203.0.113.254.4739: UDP, length 472
12:27:56.503129 Out 74:83:ef:1e:7f:99 ethertype IPv4 (0x0800), length 364: 172.24.12.3.49307 > 203.0.113.254.4739: UDP, length 320
```

After:

```
12:28:15.287022 Out 74:83:ef:1e:7f:99 ethertype IPv4 (0x0800), length 504: 203.0.113.2.51519 > 203.0.113.254.4739: UDP, length 460
12:28:15.287076 Out 74:83:ef:1e:7f:99 ethertype IPv4 (0x0800), length 516: 203.0.113.2.43431 > 203.0.113.254.4739: UDP, length 472
12:28:15.287118 Out 74:83:ef:1e:7f:99 ethertype IPv4 (0x0800), length 516: 203.0.113.2.51492 > 203.0.113.254.4739: UDP, length 472
```

### Checklist

I have:

### DIFF
--- a/src/tee_plugin/tee_plugin.c
+++ b/src/tee_plugin/tee_plugin.c
@@ -708,9 +708,14 @@ int Tee_prepare_sock(struct sockaddr *addr, socklen_t len, u_int16_t src_port, i
     memset(&source_ip, 0, sizeof(source_ip));
     memset(&ssource_ip, 0, sizeof(ssource_ip));
 
-    if (src_port) { 
-      source_ip.family = addr->sa_family; 
-      ret = addr_to_sa((struct sockaddr *) &ssource_ip, &source_ip, src_port);
+    if (config.nfprobe_source_ip) {
+      ret = str_to_addr(config.nfprobe_source_ip, &source_ip);
+      addr_to_sa((struct sockaddr *) &ssource_ip, &source_ip, src_port);
+    } else {
+      if (src_port) {
+        source_ip.family = addr->sa_family;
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Hi Jeremiah ( @floatingstatic ),

Thanks for giving a shot to this. What you did indeed makes it working (and i agree there was some overlook in the original patch, sigh). Problem being that `Tee_prepare_sock()` is called also in another part of the code where the knob `config.nfprobe_source_ip` does not apply.

So the proper way would be to add another argument to the function, say `src_ip`, and implement the logic against it. Also, one may have `src_ip` or `src_port` or both actually! So they should not be evaluated as an if-then-else but rather: if `src_ip` set it; if `src_port` set it. What is not set is then picked by the kernel.

Let me try to give it a shot later today in my day. 

Paolo

PR #503, opened 2021-05-13
Minor docker build improvements

dcaba:
### Short description

This PR just bring minor build improvements, mostly in the base Dockerfile:
* In order to reduce disk usage a bit (suggesting to move to the Debian slim flavor, disabling man pages installation, cleaning apt dirs)
* In order to improve build performance (compiling with `-j CPU_CORES`)
* to avoid surprises (`.dockerignore` pointing to `.gitignore`, that ensures we don't add to the build context/resulting images things that we normally don't commit to git, as artifacts from builds in your workstation, and pulling software checking certificates... not doing that is a sec risk)
* And, finally, minor formatting improvements (the `MAINTAINER` statement is deprecated)

A second PR will bring a multi-stage proposal as, despite the explicit package cleanup, that is partial and we are keeping lots of build requirements in the final runtime image.

### Checklist

_Note: no new files introduced, neither source code changed. If there are manual image validation steps, let me know... I've built things locally, but that's it. And if you want me to update some docs or CHANGELOG, just let me know!_

I have:
- [ ] included documentation (including possible behavior changes)


paololucente (commented):
Hola Dani ( @dcaba ),

Thank you very much for these two contributions, much-much appreciated. I will defer the review to the great Marc ( @msune ) as he has conceived and built the current incarnation of pmacct Docker and CI.

Paolo

msune:
@dcaba First, thank you for this PR. Good improvements overall. All looks good to me except for the `--no-check-certificate`.

I don't know why these are there. They came from the existing CI build scripts already. @paololucente do you recall why is that? I agree on the rule of not disabling checking certificates in general, but I guess it's worth double checking.

Aside from this, we are missing the CI run:

>1 workflow awaiting approval (First-time contributors need a maintainer to approve running workflows. Learn more.)

@paololucente could you do the honours? 

If we get a green light for the CI and `--no-check-certificate` is clarified, then it's a :+1: from me.

.dockerignore
```diff
@@ -0,0 +1 @@
+.gitignore
```
msune, line 1: See my comment here:

https://github.com/pmacct/pmacct/pull/504#issuecomment-841541647

Please clarify.

docker/base/Dockerfile
```diff
@@ -44,19 +46,19 @@ RUN apt-get update && \
...
-  sh autogen.sh && ./configure --enable-mysql --enable-pgsql     \
+  ./autogen.sh && ./configure --enable-mysql --enable-pgsql     \
                                --enable-sqlite3 --enable-kafka   \
                                --enable-geoipv2 --enable-jansson \
                                --enable-rabbitmq --enable-nflog  \
                                --enable-ndpi --enable-zmq        \
                                --enable-avro --enable-serdes     \
                                --enable-redis --enable-gnutls && \
-  sudo make install && \
+  sudo make -j$(nproc) install && \
```
msune, line 61: See my comment here https://github.com/pmacct/pmacct/pull/504#discussion_r632854940 on the same topic.

paololucente:
> --no-check-certificate is clarified

@msune : i am supportive for that one: it happened a couple of times - only couple of times in, dunno, few years but still it did happen - that certificates at an https website randomly expire (and i clearly remember one of the apache.org websites did that!) and then the CI does not complete.

msune:
> > --no-check-certificate is clarified
> 
> @msune : i am supportive for that one: it happened a couple of times - only couple of times in, dunno, few years but still it did happen - that certificates at an https website randomly expire (and i clearly remember one of the apache.org websites did that!) and then the CI does not complete.

If this is the case we have three options:

a) Leave it as is, understanding the security risks (essentially DNS attacks etc..)

b) Remove it, and run the risk of breaking pipelines. If this would happen again, a commit re-enabling it would be needed

c) In `deps.sh` check for the env variable `DEPS_DONT_CHECK_CERTIFICATE` and conditionally add `--no-check-certificate` accordingly. By default this env variable would not be defined. This has the advantage that, by just defining this variable in Github Action's env variables, the behaviour could be changed, without the need of an extra commit.

I have a strong preference for c).

paololucente:
@msune : with your C option, you indeed made a proposal that can't be refused. Support.

.dockerignore
dcaba: Hi! Yes, I also try to keep `.dockerignores` as softlinks to `.gitignore` files (as I initially attempted here), unless there are additions/extra lines that justify the copy. And while I was doing the multi-stage changes, I realized there is a line that is very beneficial (which tends to be a quite recurrent scenario) and justifies that copy: 
https://github.com/pmacct/pmacct/pull/504/files#diff-2f754321d62f08ba8392b9b168b83e24ea2852bb5d815d63e767f6c3d23c6ac5R53

Without that, as an example, a change in the last line of your `Dockerfile` invalidates all cached layers for everything after a `COPY . .`, as the Dockerfile itself changes, while docker is smart enough to just invalidate the layers behind the Dockerfile change if the Dockerfile is not in the build context.

Bringing that file change to this PR, as suggested.

docker/base/Dockerfile
dcaba: I can drop this, but I do feel that in newer CPUs activating parallelism like this makes sense, even more sense given the number of cores are increasing; without this param, unless we are calling gcc with something like `gcc -pipe`, nothing is going to parallelize compilation work, and I don't think that is our case given these tests in different containers (to avoid things being reused):

```bash
root@fbbb4bd280e4:/tmp/pmacct# time make
...
make[1]: Leaving directory '/tmp/pmacct'

real    2m54.137s
user    2m41.248s
sys     0m15.969s
...
root@6af4099cd490:/tmp/pmacct# time make -j 4
...
make[1]: Leaving directory '/tmp/pmacct'
real    1m27.542s
user    3m31.861s
sys     0m22.642s
```

I have been visualizing cpu usage in the two tests, that confirms that without `-j` we just use a single process and cpu core.

Regarding "not being polite" with this strategy, I guess these lines tends to execute in two scenarios:
a) CI system
b) developers/contributors workstations (to test changes, generate local images with local changes...)

For a), I tend to discover and pick all cores available, as CI systems normally isolate different CI jobs and give you dedicated VMs/containers with cgroups, so you can just make the best usage of the resources you see. For b), my very personal view is that I expect processes in my workstation to use as many resources as available to execute as fast or interactive as possible and if, as an user, I want to split my resources to avoid things affecting to each other (better browsing experience while I'm executing heavy work as a compilation process?), it's my job to setup cgroups or other isolation techniques. 

Anyways, if you want a build ARG that defaults to this, but users can override at will, I can work on that. I would not recommend to hardcode the current CI workers specs (they may change, and you are ignoring the specs of users/contributors workstations)

dcaba:
> @msune : with your C option, you indeed made a proposal that can't be refused. Support.

I'm implementing C, but I really think shipping software (and dependencies) from sources we cannot verify is just something that should not be done, you can be distributing malicious software. These checks are there because of something; it's like unit tests execution: if one fails, it means something, and you cannot just skip CI tests execution in order to ship a new version. If you ever find yourself in that situation, I would recommend you to even vendor your dependency (by manually pulling it, verify signatures, and do your best to ensure no-one is taking advantage of this cert error to inject malicious stuff) rather than making CI to trust unverified sources (it's not just about DNS attacks: you know better than me that IP traffic can be hijacked also with BGP announcements, for instance)

.dockerignore
msune: >Without that, as an example, a change in the last line of your Dockerfile invalidates all cached layers 

I see make sense.

To avoid having to maintain two times the same file (the contents of `.gitignore`), we could create `.dockerignore-base` containing just the extra lines to add on top of `.gitignore`. We would still track `.dockerignore`.

In this way, if a change is done in `.gitignore`, you can easily do:

```
cat .gitignore > .dockerignore
cat .dockerignore-base >> .dockerignore
```

Even the pipeline can have a simple check to make sure `.gitignore` is actually container in the tracked version of `.dockerignore` to avoid getting desynced. Thoughts?

docker/base/Dockerfile
msune: >For a), I tend to discover and pick all cores available, as CI systems normally isolate different CI jobs and give you dedicated VMs/containers with cgroups, so you can just make the best usage of the resources you see. 

Using `-j$(nproc)` is a **very bad** idea for _any_ of the options listed.

First you never want to have `make -j $(nproc)` be the number of lcores, even if you are a single user of the machine/vm/whatever. _If_ a `Makefile` has enough translation units (>= nproc) you can severly impact the performance of the machine. Not sure if you ever have done `make -j8` in an 8 lcore machine, but I can tell you I did, and with intensive compilations, and even I/O (mouse, keyboard) of my laptop were laggish, to the point of being unusable. At most you want `nproc-1` to be your workers, assuming this is your only real job.

Second, many (most nowadays) of the times systems are shared; shared servers, VMs, containers. For instance for docker containers, the default is to **NOT limit** the number of CPUs "seen" by the container, unless you explicitely set it:

https://docs.docker.com/config/containers/resource_constraints/#cpu

```
By default, each container’s access to the host machine’s CPU cycles is unlimited.
```

This means that they _see_ the lcores of the host. You can imagine the consequences with more than one container - which is 99% the case in servers hosting docker containers.

See:

```
marc@Mriya:~$ docker run -it debian:buster-slim
root@4faac422c41f:/# nproc 
8
root@4faac422c41f:/# exit
marc@Mriya:~$ docker run -it --cpus=3 debian:buster-slim
root@7becf73d2ed8:/# nproc
8
root@7becf73d2ed8:/# exit
marc@Mriya:~$ docker run -it --cpuset-cpus=1,2 debian:buster-slim
root@f073cb48d1cd:/# nproc
2
root@f073cb48d1cd:/# 
```

The only way to limit `nproc` is to actually use  `--cpuset-cpus`, which is _very_ inconvenient, because that's _the_ list of  lcores (pinning) and not an lcore counter (`--cpus` just computes the equivalent CPU quota, but `nproc` is unaltered). So if you need to "limit" `nproc` you need to start load balancing lcores via `--cpuset-cpus` for all containers, which, again, is _very_ inconvenient. 


If you are still not convinced, I invite you to google a bit on the problems of using `nproc` as an arg to `make -j`. First google result, see the replies:

```
https://github.com/OpenVisualCloud/Dockerfiles/issues/19
```

> Anyways, if you want a build ARG that defaults to this, but users can override at will, I can work on that.

I am advocating for an environament variable _not_ a script argument. If not set define it to a reasonable default (e.g. 2) at the very beginning of the script:

```
if test -z $NUM_WORKERS; then
  NUM_WORKERS=2
fi

#...

#Wherever make is invoked
make -j  $NUM_WORKERS
```

dcaba: Hi Marc. Thanks for the elaborated response:

* Yes, i tend to match `-j` with the number of cores in my system, and I cannot have a VC meeting in parallel, but I don't experience the degradations you are describing (different story is when the `-j` arg exceeds the number of cores in your system, that generates excessive context switching and increases queues). 

* Despite modern servers do have a big number of cpus, I don't think that means we should stick to `1` core, or the fact that increasing this does not come with advantages, which is what I was representing with the previous examples.

* And yes, you are right: I forgot for a second about the complexity of identifying the effective CPU shares you can use inside cgroups, and the fact you see the host cpus, regardless the cpu limits you may be defining. I need to argue, though, that despite most of the CI systems use shared infrastructure, it's strange you land in a non dedicated VM, which gives you the subset of the system you normally are able to consume, but not more. Or, if there are implementations with shared virtual CPUs (I doubt that, given the security implications) and just cgroups for resource management, cpu limits are going to prevent you from doing an unfair usage of the system (a high -j value may mean throttling, to force you to stay under the limits, at the cost of perf degradation compared to a number of parallel jobs that matches the cpu limits for your partition)

* Anyways, regarding the support of a configurable `NUM_WORKERS` param which defaults to 2, I was thinking about a docker-build arg, given the make install lives at the `Dockerfile` level. And we can propagate that to the `ci/deps.sh` script, in the parallel PR

msune:
> I'm implementing C, but I really think shipping software (and dependencies) from sources we cannot verify is just something that should not be done, you can be distributing malicious software. 

No one is arguing against this, I think. I am with you on this.

>These checks are there because of something; it's like unit tests execution: if one fails, it means something, and you cannot just skip CI tests execution in order to ship a new version. If you ever find yourself in that situation, I would recommend you to even vendor your dependency (by manually pulling it, verify signatures, and do your best to ensure no-one is taking advantage of this cert error to inject malicious stuff) rather than making CI to trust unverified sources (it's not just about DNS attacks: you know better than me that IP traffic can be hijacked also with BGP announcements, for instance)

Yes, I think everyone here is aware of it. I am not sure you are aware/understanding the trade offs and consequences here of _not_ doing option C (everything has a trade-off :wink: ), so let me briefly comment.

First and foremost; yes there are attacks, but the most common cause for that check to fail is an expired cert, so really a bad sysadmin somewhere in apache foundation (according to Paolo's), not an attack. See more below on the consequences of a true BGP hijack.

Second, this script is not used by users to build pmacct. This is _only_ used by:

* Build tests 
* Docker images

Users compiling pmacct from the sources have to download and install  dependencies, so it's up to them to do any security assessment.

Third, on pmacct's official Docker containers...

I furiously agree with you that we should NEVER release with `DEPS_DONT_CHECK_CERTIFICATE=1`. We can set this as a rule @paololucente.

There is still the case of `bleeding-edge` tag, which is distributed on every push to `master` (with green pipeline).

As said before, `DEPS_DONT_CHECK_CERTIFICATE`  should never be defined, just during exceptional times, and this is the proposal. If some of the issues listed above would happen, with option C the pipeline will fail... RED.

It's still under the control of @paololucente to see _why_ the pipeline failed, and assess on which exact case we are. @paololucente should diligently decide if it's safe to enable `DEPS_DONT_CHECK_CERTIFICATE` or not (perhaps by downloading the sw from a container in GA nad verifying md5sum etc..) or stop because there is really a security concern.

This is no different than your proposal (B)... 

> Just a minor observation; if someone _truly_ BGP hijacks some prefixes containing web servers for the deps in the script (e.g. apache), github servers will be affected by that. So we are going to talk on a massive security issue that will for sure be top priority to be solved

I see a risk of leaving that setting on beyond the problem, but again I am sure @paololucente will be diligent on this.

Now the downside of your proposal. All that you have said, including mirroring deps (see below for this) etc.. _as a result_ of that check failed is a substantial amount of work to be done and will severely block any PR and commits of Paolo.

Since the most probably cause is a temporary error like a sysadmin not having renewed a cert (or not having uploaded it correctly...), it's possible the hardcoded `--no-check-certificate` will be integrated again, which is far worst than having C temporally enabled. The community needs to continue to build, integrate patches etc... If one of the libs (plugins) is _really_ in serious trouble because it has been hacked, no worries that as part of the assessment it will be either disabled from the docker bleeding-edge image or temporally removed. 

>I would recommend you to even vendor your dependency (by manually pulling it, verify signatures, and do your best to ensure no-one is taking advantage of this cert error to inject malicious stuff)

For the record, I've been proposing to @paololucente  for sometime to mirror pmacct deps in a controlled env, and use submodules. This would simplify the process of building pmacct from the sources, and we will have more control. This would completely avoid the  problem you are mentioning. In fact we did it already for libcdada this way, as a proof-of-concept (except for the part of mirroring the dep (still submodule SHA ensures it's the same code or will fail to checkout if commit is gone).

.dockerignore
dcaba: Despite I do agree maintaining the duplication is not great (and I tend to avoid things like this as much as possible), I feel the associated overhead is already "assumed" by the community, so this special `.dockerignore` management may be a bit surprising for contributors. But we can implement that check, and maybe consider an alternative: CI checking that the `.dockerignore` contains everything we have in `.gitignore` plus a set of lines from a whitelist (initially just the `Dockerfile` line),  which is the property we are looking for, so we avoid the introduction of that `dockerignore-base` and we reduce the surprise factor.

docker/base/Dockerfile
msune: > Despite modern servers do have a big number of cpus, I don't think that means we should stick to 1 core, or the fact that increasing this does not come with advantages, which is what I was representing with the previous examples.

Absolutely, and that's why the contrib is useful (thanks again!). To be honest I didn't even check that script, as it was already done for the build CI.

If you want:

* We can set `NUM_WORKERS` by default to 4, which in most workstations and laptops is a sane value nowadays for building dockers locally, and close to the "maximum" real usable `-j` value (I suspect it's 6 or 8 based on my experience in the laptop) 
* We can set the fine tuned  `NUM_WORKERS`  for CI in the Github Actions env.

> cpu limits are going to prevent you from doing an unfair usage of the system (a high -j value may mean throttling, to force you to stay under the limits, at the cost of perf degradation compared to a number of parallel jobs that matches the cpu limits for your partition)

Exactly, which somehow defeats the purpose.

.dockerignore
msune: That would be my preference, but let's see what @paololucente thinks. If the decision is to duplicate, then... well, I will learn to live with it :joy:

docker/base/Dockerfile
msune: >Anyways, regarding the support of a configurable NUM_WORKERS param which defaults to 2, I was thinking about a docker-build arg, given the make install lives at the Dockerfile level. And we can propagate that to the ci/deps.sh script, in the parallel PR

I forgot to answer this one... That'd be sort of the recommended way to do it in `docker`, I agree. If three is an easy way to have a reasonable default when not specified sure, let's go that way. We can invoke it with that build option in the CI (as we want to have it as a Github Actions env variable no matter what)

msune:
@paololucente I think @dcaba is still working on the PR, so please don't merge just yet.

@dcaba whenver you are done, please ping me to have another review.

dcaba:
> @paololucente I think @dcaba is still working on the PR, so please don't merge just yet.
> 
> @dcaba whenver you are done, please ping me to have another review.

Hi @msune : I think you can review it again: 
* `NUM_WORKERS` with a default of 2 has been declared in the base `Dockerfile`
* I'm expecting to cover the parallelism of the deps installation in the next PR, that is where it was initially introduced, honoring the previous setting
* cert-checks as a configurable variable (that inherits its value from an optional env var) has been introduced in the `ci/deps.sh` script

So, unless I'm missing something, hopefully we are ready to go and focus in the next one.

ci/deps.sh
```diff
@@ -20,9 +20,14 @@
 #Do not delete
 set -e
 
-#wget Retries
+#wget options
 WGET_N_RETRIES=30
 WGET_WAIT_RETRIES_S=10
+WGET_FLAGS="-t $WGET_N_RETRIES --waitretry=$WGET_WAIT_RETRIES_S"
+DEPS_DONT_CHECK_CERTIFICATE="${DEPS_DONT_CHECK_CERTIFICATE:-}"
```
msune, line 27: Hopefully the last comment. Seems this doesn't work without `bash`/`dash` (albeit it worked in `ksh`

With `tcsh` it doesn't, so I suspect it's not "safe" (`sh` portability):

```
root@9f5bc224f6b2:/# cat test.sh 
#!/bin/sh
TEST=" ffff "
TEST2=${TEST:-}
echo "$TEST2"
root@9f5bc224f6b2:/# rm /bin/sh 
root@9f5bc224f6b2:/# ln -s /bin/tcsh /bin/sh
root@9f5bc224f6b2:/# sh test.sh 
TEST= ffff : Command not found.
Bad : modifier in $ '-'.
```

dcaba: Good catch. Yes, I'm maybe too used to bash functionality, and I don't think defining vars with defaults like this is supported in many other shells. However, that line is redundant (as taking the value of a var, or an empty value as a default, is almost the same than just picking the env var directly), but I tend to incorporate this even in this scenario, to make that env var value recollection/usage explicit and clear at the beginning of the script (rather than finding out the usage of custom env vars in the middle of scripts) plus, if my memory is not wrong, that also lets you use the "bash nounset mode" (`-u`) in case the env var is not defined.

So I'm going to just drop that line: I guess the script is not long, and the value of `DEPS_DONT_CHECK_CERTIFICATE` is picked quite early in the script, so an user of this will quickly identify that functionality (if we ever want to). But, anyways, my very personal opinion about these topics with this script (and it's an opinion from someone with very little context, so it has relative value): I think the cost of trying to maintain an script that installs deps in very diverse environments is quite high... I would consider (if possible) that, for users needing to install pmacct in a system without bash, let the person interested on that to care about deps installation, or to install bash in their system, rather than trying to consider this an "universal script" and assume the extra maintenance cost. And if you really want to keep property, maybe CI can do these compat checks for you (for instance, in some software/scripts I maintain which are expected to be compat with osx, I tend to run them on top of some deprecated base images that come with ancient versions of bash (ubuntu precise?), mimicking osx versioning... maybe you can find or create some base images that can represent your target environments)

msune:
(sorry for the late reply, but has been a busy day and Debian servers are ridiculously slow for me today, so validation took me a lot)

I've tried it locally. I've added some prints locally in `deps.sh` to see (`NUM_WORKERS` and `DEPS_DONT_CHECK_CERTIFICATE`) , might make sense adding them for real. 

`NUM_WORKERS` works fine, but  `DEPS_DONT_CHECK_CERTIFICATE` doesn't seem to be set as `ARG` (or `ENV` for that matter) in the `Dockerfile` of the base container, so obviously `--build-arg DEPS_DONT_CHECK_CERTIFICATE=1` doesn't work as expected.

Btw, (sorry for not having spotted this before) the CI code is not passing any of these variables - env variables set by github actions (`NUM_WORKERS` and `DEPS_DONT_CHECK_CERTIFICATE`) - via `--build-arg` to `docker build` for the base container, which is the only one that needs them:

https://github.com/dcaba/pmacct/blob/docker-build-improvements/.github/workflows/ci.yaml#L149

Might be worth fixing it here too.

msune:
Paolo suggested to post this here, since I have it locally (I am sure you don't need it :wink: ):

```
marc@Mriya:~/personal/pmacct$ git diff
diff --git a/ci/deps.sh b/ci/deps.sh
index 72811bb8..3240afb3 100755
--- a/ci/deps.sh
+++ b/ci/deps.sh
@@ -28,6 +28,9 @@ if [ "${DEPS_DONT_CHECK_CERTIFICATE}" ]; then
     WGET_FLAGS="${WGET_FLAGS} --no-check-certificate"
 fi
+echo "WGET_FLAGS: ${WGET_FLAGS}"
+echo "NUM_WORKERS: ${NUM_WORKERS}"
+
 #Don't pollute /
 mkdir -p /tmp
 cd /tmp
diff --git a/docker/base/Dockerfile b/docker/base/Dockerfile
index 3cf0e222..dd92919d 100644
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -18,6 +18,8 @@ COPY . /tmp/pmacct/
 #Parallelism: 2 looks a reasonable default, and bear in mind CI specs
 ARG NUM_WORKERS=2

+#Do not check certificates in wget for external deps
+ARG DEPS_DONT_CHECK_CERTIFICATE
+
 #Create a single compressed layer, as some dependencies (deps.sh)
 #can be later removed, so try to make it as lightweight as possible
 RUN apt-get update && \
marc@Mriya:~/personal/pmacct$
```

The line in the `ci.yml`:

```
docker build --build-arg NUM_WORKERS=$CI_NUM_WORKERS --build-arg DEPS_DONT_CHECK_CERTIFICATE=$CI_DEPS_DONT_CHECK_CERTIFICATE -f docker/base/Dockerfile -t base:_build
```

ci/deps.sh
msune: > would consider (if possible) that, for users needing to install pmacct in a system without bash, let the person interested on that to care about deps installation, or to install bash in their system, rather than trying to consider this an "universal script" and assume the extra maintenance cost. 

This script is only used by the CI a) build checks b) docker

dcaba: Got it. So my very personal recommendation: current CI coverage then wouldn't break by using bash, and if you ever have more "build shapes" / release this software as other artifact types with more builders/base images, I guess bash could be also installed in these other Unix-like systems, so I would consider just embracing bash(4?). But, in any case, that was an almost off-topic conversation: we are back to sh-compat, so marking this as resolved.

dcaba:
Hi @msune . I brought your diff to this branch, but not injecting in the `deps.ci` script `NUM_WORKERS` and echoing that, as this was planned to be part of the next PR (https://github.com/pmacct/pmacct/pull/504/files#r633555182 - given we started some discussions there, and other improvements in the same script, I didnt want to bring more noise here). Let me know if that is ok and I can rebase the parallel one with these approved+merged changes. And dont worry about the delay in the responses: this is OSS, I dont expect immediate & 24/7 availability.

msune:
@paololucente can you enable the CI for this last commit? This is annoying...

If pipeline is green, ACK. Please merge.

@dcaba since you will work on 504, I think this is superflous https://github.com/pmacct/pmacct/pull/503/commits/84557360e65e7628cfc0204d76b8c6dac64e34e6#diff-a7a92e793e13161b58a26c6c1c0445fef52d1b6fd0cc081d71701d0a1ad5c089R56

PR #504, opened 2021-05-13
Base docker build using multistage, which makes runtime more efficient and secure

dcaba:
### Short description

This PR introduces more build related improvements (so it continues the work we started here: https://github.com/pmacct/pmacct/pull/503):

* we have incorporated an stage in the CI logic / actions that builds in parallel the base image we use for dockerhub, so if we break something in that specific build flavor, we are not going to detect this when merging to master, that is when this specific docker flow is used to publish images to dockerhub.
* in the `ci/deps.sh` we stop using "&&" to concatenate installation commands, as one failing in the middle was not breaking the build process. We also enabled parallelism when building these deps, to improve performance
* And major changes in the base `Dockerfile`:
  * We move to multistage, so we don't need to clean build related deps and we just explicitly declare the ones that are required at runtime
  * In the build stage, we stop installing everything in a single layer, to improve cacheability, and that does not affect to the final/runtime stage anymore
  * With the definition of the previous introduction of `.dockerignore`, we no longer need to call `maintainer-clean`, as files not under version control are excluded.

That results in a base image that is around 300MB, 150MB less than what we already won in the previous batch of improvements. And it's not just about size (that is good for the planet): less runtime software means less chances of security issues.

### Checklist

_Note: no new files introduced, neither source code changed. If there are manual image validation steps, let me know... I've built things locally, already `ldd`'ed most of the assets in the resulting image `/usr/local`, and I didn't find anything linked to non existing libs. And if you want me to update some docs or CHANGELOG, just let me know!_

I have:
- [] added the [LICENSE template](https://github.com/pmacct/pmacct/blob/master/LICENSE.template) to new files
- [] compiled & tested this code
- [] included documentation (including possible behaviour changes)

### DIFF
--- a/ci/deps.sh
+++ b/ci/deps.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 #    Copyright
 #    (c) 2020 Marc Sune <marcdevel@gmail.com>
@@ -18,7 +18,7 @@
 #    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 # Do not delete
-set -euo pipefail
+set -e
 
 #wget options
 WGET_N_RETRIES=30
@@ -28,13 +28,11 @@ if [ "${DEPS_DONT_CHECK_CERTIFICATE}" ]; then
     WGET_FLAGS="${WGET_FLAGS} --no-check-certificate"
 fi
 echo "WGET_FLAGS: ${WGET_FLAGS}"
+echo "MAKEFLAGS: ${MAKEFLAGS}"
 
 # Don't pollute /
 mkdir -p /tmp
 cd /tmp
-
-# Most of the software installation will probably honor this, which will improve perf
-export MAKEFLAGS="-j$(nproc)"
 
 # Dependencies (not fulfilled by Dockerfile)
 git clone https://github.com/akheron/jansson
--- a/ci/deps.sh
+++ b/ci/deps.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 #    Copyright
 #    (c) 2020 Marc Sune <marcdevel@gmail.com>
@@ -17,8 +17,8 @@
 #    ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 #    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-#Do not delete
-set -e
+# Do not delete
+set -euo pipefail
 
 #wget options
 WGET_N_RETRIES=30
@@ -29,41 +29,44 @@ if [ "${DEPS_DONT_CHECK_CERTIFICATE}" ]; then
 fi
 echo "WGET_FLAGS: ${WGET_FLAGS}"
 
-#Don't pollute /
+# Don't pollute /
 mkdir -p /tmp
 cd /tmp
 
-#Dependencies (not fulfilled by Dockerfile)
+# Most of the software installation will probably honor this, which will improve perf
+export MAKEFLAGS="-j$(nproc)"
+
+# Dependencies (not fulfilled by Dockerfile)
 git clone https://github.com/akheron/jansson
-cd jansson && rm -rf ./.git && autoreconf -i && ./configure && make && sudo make install && cd ..
+cd jansson ; rm -rf ./.git ; autoreconf -i ; ./configure ; make ; sudo make install ; cd ..
 
 git clone https://github.com/edenhill/librdkafka
-cd librdkafka && rm -rf ./.git && ./configure && make && sudo make install && cd ..
+cd librdkafka ; rm -rf ./.git ; ./configure ; make ; sudo make install ; cd ..
 
 wget ${WGET_FLAGS} https://github.com/alanxz/rabbitmq-c/archive/refs/tags/v0.11.0.tar.gz
 mv v0.11.0.tar.gz rabbitmq-c-0.11.0.tar.gz
 tar xfz rabbitmq-c-0.11.0.tar.gz
-cd rabbitmq-c-0.11.0 && rm -rf ./.git && mkdir build && cd build && cmake -DCMAKE_INSTALL_LIBDIR=lib .. && sudo cmake --build . --target install && cd .. && cd ..
+cd rabbitmq-c-0.11.0 ; rm -rf ./.git ; mkdir build ; cd build ; cmake -DCMAKE_INSTALL_LIBDIR=lib .. ; sudo cmake --build . --target install ; cd .. ; cd ..
 
 git clone --recursive https://github.com/maxmind/libmaxminddb
-cd libmaxminddb && rm -rf ./.git && ./bootstrap && ./configure && make && sudo make install && cd ..
+cd libmaxminddb ; rm -rf ./.git ; ./bootstrap ; ./configure ; make ; sudo make install ; cd ..
 
 git clone -b 3.4-stable https://github.com/ntop/nDPI
-cd nDPI && rm -rf ./.git && ./autogen.sh && ./configure && make && sudo make install && sudo ldconfig && cd ..
+cd nDPI ; rm -rf ./.git ; ./autogen.sh ; ./configure ; make ; sudo make install ; sudo ldconfig ; cd ..
 
 wget ${WGET_FLAGS} https://github.com/zeromq/libzmq/releases/download/v4.3.2/zeromq-4.3.2.tar.gz
 tar xfz zeromq-4.3.2.tar.gz
-cd zeromq-4.3.2 && ./configure && make && sudo make install && cd ..
+cd zeromq-4.3.2 ; ./configure ; make ; sudo make install ; cd ..
 
 wget ${WGET_FLAGS} https://archive.apache.org/dist/avro/avro-1.9.2/c/avro-c-1.9.2.tar.gz
 tar xfz avro-c-1.9.2.tar.gz
-cd avro-c-1.9.2 && mkdir build && cd build && cmake .. && make && sudo make install && cd .. && cd ..
+cd avro-c-1.9.2 ; mkdir build ; cd build ; cmake .. ; make ; sudo make install ; cd .. ; cd ..
 
 git clone https://github.com/confluentinc/libserdes
-cd libserdes && rm -rf ./.git && ./configure && make && sudo make install && cd ..
+cd libserdes ; rm -rf ./.git ; ./configure ; make ; sudo make install ; cd ..
 
 git clone https://github.com/redis/hiredis
-cd hiredis && rm -rf ./.git && make && sudo make install && cd ..
+cd hiredis ; rm -rf ./.git ; make ; sudo make install ; cd ..
 
-#Make sure dynamic linker is up-to-date
+# Make sure dynamic linker is up-to-date
 ldconfig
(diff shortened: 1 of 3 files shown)


paololucente (commented):
Hola Dani ( @dcaba ),

Thank you very much for these two contributions, much-much appreciated. I will defer the review to the great Marc ( @msune ) as he has conceived and built the current incarnation of pmacct Docker and CI.

Paolo

msune:
Hi @dcaba. I will have a look into this PR in detail after #503. however two comments:

> dockerignore adds the Dockerfiles and .git in this repo, to make the most of the build cache

If I read changes correctly, here you are directly (manually) copying the contents of `.gitignore` in `.dockerignore`. I would like to avoid that (symlink or simply generated by the CI itself).

I would also suggest to move this to #503, and focus this PR on multi-stange. 

Btw, I suspect (after seeing this PR) that `.dockerignore` in #503 is not doing what was intended to be done, but I didn't try it myself.

> and we stop using "&&" to concatenate installation commands, **as one failing in the middle was not breaking the build process**

I am confused... 

```
marc@Mriya:~$ /bin/true && /bin/false && /bin/true
marc@Mriya:~$ echo $?
1
marc@Mriya:~$ /bin/true && /bin/true && /bin/true
marc@Mriya:~$ echo $?
0
```

To the best of my knowledge operator `&&` has always worked this way in all shell incarnations. Could you clarify what was actually breaking the build?

ci/deps.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
msune, line 1: `pmacct` is compiled in several flavours of BSD. Since the `Dockerfile` can be used from these systems, and hopefully sometime the CI will add coverage, please do not move to `bash`.

ci/deps.sh
```diff
@@ -17,48 +17,51 @@
 #    ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 #    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-#Do not delete
-set -e
+# Do not delete
+set -euo pipefail
```
msune, line 21: idem

ci/deps.sh
```diff
@@ -17,48 +17,51 @@
...
 WGET_WAIT_RETRIES_S=10
 
-#Don't pollute /
+# Don't pollute /
 mkdir -p /tmp
 cd /tmp
 
-#Dependencies (not fulfilled by Dockerfile)
+# Most of the software installation will probably honor this, which will improve perf
+export MAKEFLAGS="-j$(nproc)"
```
msune, line 32: * Although `$()` is POSIX, but since we still want to be `sh` compatible to avoid any _possible_ portability issues - JIC - please use ``
* It's rather impolite from us to take _all_ available lcores on the system for compilation, and really unnecessary for almost all modern CPUs. Besides, it can have unintended consequences if some parts would actually use _all_ of those. I am fine on adding parallelism, but a) should have reasonable defaults (e.g. 4 or whatever) b) having a way to override via env variable.

msune: >have reasonable defaults (e.g. 4 or whatever) 

On this, I was trying to research on what is the optimal `value` in `make -j <value>` in Github Actions. I know for Travis it was in the range of 2-4, but actually results were all similar, so we were using 2. We should probably start with 2 also.

ci/deps.sh
dcaba: No probl: unaware of that requirement. I will rollback when we are happy to merge https://github.com/pmacct/pmacct/pull/503 , as this will require rebase

ci/deps.sh
dcaba: Same: yes, I will rollback after https://github.com/pmacct/pmacct/pull/503

ci/deps.sh
dcaba: As commented in parallel, we will change this and we will honor a `NUM_WORKERS` env var which default will be declared in the `Dockerfile`

dcaba:
> To the best of my knowledge operator `&&` has always worked this way in all shell incarnations. Could you clarify what was actually breaking the build?

Better an example than me trying to provide an explanation:

```bash
 $ bash -e -c "/bin/true && /bin/false && /bin/true; echo I did finish"
I did finish
 $ bash -e -c "/bin/true; /bin/false; /bin/true; echo I did finish"
 $ 
```

So the expectation of your shell is that, if you are using `&&`, you are doing error handling by your own, and `-e` will not interrupt the script execution. While if you just execute the different statements in sequence, an error will interrupt the flow, that is what we want to avoid continuing with an intermediate error. I reproduced that bug when I tried to remove the "sudo" binary from the docker image: the deps.sh script continues and then the main make fails because of something was not properly installed.

dcaba:
_Converted to draft, as its going to be rebased and refined a bit. I will mark as ready once it has been revisited_

ci/deps.sh
dcaba: Keeping sh. See last commit ; resolving

ci/deps.sh
dcaba: keeping sh compat. Resolving

ci/deps.sh
dcaba: This has been rewritten in the parallel PR. Resolving

dcaba:
Hi! This has been revisited and rebased, so it no longer conflicts with old implementations of the previous PR. PR description has been updated: as you can read there, I also took the freedom to add an extra stage to the CI, so the base Dockerfile you use to push to dockerhub is also built in PR jobs (other-ways, I think you were only stressing that flow when merging to master). Marking as ready to review again.

msune:
>Better an example than me trying to provide an explanation:

Humm... yeah. I've stumbled across this, and it's annoying. I don't quite follow the rationale on why that fails. 

To be honest I don't know why these `wget && autogen && configure..` were being done, but I suspect someone just copy pasted the commands he/he was using.

I am fine replacing them with `;`, but we could very well just have it in different lines, to avoid lines getting very long (and probably put some breaklines between deps to have a a visual grouping), but really this is a matter of taste, and it's perfectly fine as is. 

I miss Gitlab's comment threads here... sigh

msune:
Hi Daniel,

First, thank you (again) for this contribution. The space saved is massive, so kudos!

A couple of general comments:

* First, very nice using (or keep using) `/tmp` with the multi-stage, as the cleanup is done there for free.
* I think multi-stage is a nice concept (and a pity it wasn't there sooner in docker). The code looks cleaner when using it. For pmacct in particular, however, I am not sure it's useful beyond that. All daemons just change the entrypoint from base container, as the SW package is compiled "all at once". So unless I am missing something, there is no "other gain". I am wondering if you were thinking in some follow-up contribs that would make more use of it.
* If we would ever get some benefit from multi-stage, we might as well consider a 4-tier (docker) stage: 1) apt-get 2) ci-deps 3) pmacct build 4) cleanup. There is absolutely no sense on doing all this work now given my comment above.

Issues:

a) First and foremost, I didn't run it in my user's env, but I think the docker-hub stage is broken. Registries are not kept between jobs, so you need to export them as artifact, else `docker-hub` job won't find docker container `base`. See: 

https://github.com/pmacct/pmacct/blob/master/.github/workflows/ci.yaml#L33
https://github.com/pmacct/pmacct/blob/master/.github/workflows/ci.yaml#L36

Comments:

- (Something I was just too lazy to do it in the initial docker effort) We should probably split in two different stages the docker container (pmacct) generation from docker-hub upload. In this way PRs would only skip upload (for sec. reasons), and we would see the failures in the docker generation in the PR itself - now it's skipped.
- I recommend to test docker stage in your fork (you need to enable GA). The upload is a bit painful, as you have to create all these dockerhub repos... ping me if you want me to test it, as a I have a mirror of them. But you should see it reaches the upload part of the CI
- I don't quite understand the moving of base container, see below

b) Is there a reason why the new job `builder-base-docker` is created and runs _before_ the `build-and-test`? The thing is, the  `builder-*` stage (btw we should put a name on it) was meant to generate the auxiliary containers to build pmacct (hence the name builder). The base container is actually having already pmacct compiled in, which is a bit different. In fact, it was intentional to compile the base container at the docker-hub stage, as if the tests were not passing in the stage before, there is really no point (waste of cycles) on compiling the base container, as it's only used for the docker images.

c) Setting aside the comment above, in the current PR's proposed workflow `docker-hub` stage job needs to depend on `builder-base-docker`. It's highly improbable, but it's a bug in the pipeline definition.

dcaba:
Hi @msune ! I'm sorry, as I think the last-minute CI change is adding more noise than noise: my aim with that extra stage was just to execute a docker-build against the base image also in non-master build jobs, as if someone breaks the base image build (it could be me with this PR, for instance), you are not going to realize of this until you merge the contribution, as the base image is only used in that publish-to-github flow. But yes, in order to make this better / reuse work in the master job, we should plug this to the `docker-hub` step, rather than being something independent which duplicates a portion of the job. Not a Github Actions expert, very very basic user, but I can help with that if you need (but I would prefer to keep that outside of this PR, as yes, it will imply testing the entire flow in my fork VS this harmless change)

So the main contribution here is the multistage transition, and the fix in the deps script that means the build process is not going to stop in some failure scenarios. And, just with it, we are delivering a cleaner and more slim docker image, compared to the previous approach (450MB vs 300MB, uncompressed, and by far, less packages installed in the resulting image), as the "pmacct" clean-up stage was not comprehensive, and yes, I find this easier to maintain than needing to care of removing and cleaning each compile-time dependency we may ever need.

msune:
OK. Please revert/remove commits that changes the CI, and let's stick to the multi-stage contribution here.

Please notify me when ready, and I will have a final (and manual) check for Dockers

msune:
Split docker in two different jobs in #506., which should probably be integrated before this PR to make sure we run docker image generation cc @paololucente

dcaba:
> OK. Please revert/remove commits that changes the CI, and let's stick to the multi-stage contribution here.
> 
> Please notify me when ready, and I will have a final (and manual) check for Dockers

Done, @msune !

msune:
Glanced through it. I won't retest, I trust @dcaba. ACK @paololucente

dcaba:
@paololucente , do you plan to release a new version soon? Thanks to the improvements we have introduced, the resulting docker images are prettier.. see compressed size:

image

and number of sec findings:

```bash
 $ trivy --quiet pmacct/sfacctd:v1.7.6 | head

pmacct/sfacctd:v1.7.6 (debian 10.8)
===================================
Total: 406 (UNKNOWN: 0, LOW: 193, MEDIUM: 93, HIGH: 103, CRITICAL: 17)

$ trivy --quiet pmacct/sfacctd:bleeding-edge | head

pmacct/sfacctd:bleeding-edge (debian 10.9)
==========================================
Total: 137 (UNKNOWN: 0, LOW: 95, MEDIUM: 15, HIGH: 25, CRITICAL: 2)
```

So despite there may not be major code updates, I think the community can benefit from this.

paololucente:
Hi Dani ( @dcaba ),

I am wrapping up one important feature before code freeze. It will maybe take a week or so to be completely done. So sometime in July (first half / mid) seems realistic for the new release.

Let me take this chance to repeat how grateful i am for your contributions, thanks very much.

Paolo

dcaba:
> Hi Dani ( @dcaba ),
> 
> I am wrapping up one important feature before code freeze. It will maybe take a week or so to be completely done. So sometime in July (first half / mid) seems realistic for the new release.
> 
> Let me take this chance to repeat how grateful i am for your contributions, thanks very much.
> 
> Paolo

Got it, @paololucente , thanks for sharing these plans!

And thanks also for maintaining this project! A pleasure to help and discuss with @msune : Marc, if you ever need some fresh eyes to review some build-related stuff in this repo, feel free to mention or assign me the review role.

msune:
Daniel, likewise. I will sure put you in the loop in future contribs

Thanks!

PR #506, opened 2021-05-25
ci: split docker and dockerhub upload jobs

msune:
### Short description

Prior to this commit both the generation of docker images and
uploading them to dockerhub was done in the same job. This meant
that docker generation was skipped for all PRs.

In order to solve this, this commit splits this in two sequential
jobs, one to compile the docker images (and exporting the registry
as artifact) and one publishing them to dockerhub (having imported
the registry artifact). The former will also run for PRs.

### Checklist
I have:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -126,7 +126,6 @@ jobs:
 
   ### Step 3: official docker image generation
   pmacct-docker:
-    needs: build-and-test
     runs-on: ubuntu-18.04
     steps:
       - name: Checkout pmacct
@@ -165,7 +164,7 @@ jobs:
 
   ### Step 4: Upload images to dockerhub (bleeding-edge, latest and releases)
   publish-dockerhub:
-    needs: pmacct-docker
+    needs: [pmacct-docker, build-and-test]
     runs-on: ubuntu-18.04
     if: github.event_name != 'pull_request'
     env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -124,21 +124,17 @@ jobs:
           docker exec -i $CONTAINER_ID ./ci/script.sh
           docker stop $CONTAINER_ID
 
-  ### Step 3: official dockerhub images (bleeding-edge, latest and releases)
-  docker-hub:
+  ### Step 3: official docker image generation
+  pmacct-docker:
     needs: build-and-test
     runs-on: ubuntu-18.04
-    if: github.event_name != 'pull_request'
-    env:
-      DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-      DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
     steps:
       - name: Checkout pmacct
         uses: actions/checkout@v1 #Don't use v2 messes everything
         with:
           path: pmacct
 
-      - name: Build and upload containers
+      - name: Build containers
         run: |
           echo "Fix mess with tags in actions/checkout..."
           git fetch -f && git fetch -f --tags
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -124,21 +124,17 @@ jobs:
           docker exec -i $CONTAINER_ID ./ci/script.sh
           docker stop $CONTAINER_ID
 
-  ### Step 3: official dockerhub images (bleeding-edge, latest and releases)
-  docker-hub:
+  ### Step 3: official docker image generation
+  pmacct-docker:
     needs: build-and-test
```
dcaba, line 129: Do you really want to keep that dependency? I think this stage is something you can do in parallel to testing images and execution, so it can be triggered at the very beginning of this pipeline. But then publish-dockerhub depending on these images generation and the successful tests execution (this build-and-test)

.github/workflows/ci.yaml
```diff
@@ -155,6 +151,48 @@ jobs:
           docker build -f docker/pmbgpd/Dockerfile -t pmbgpd:_build .
```
dcaba, line 151: maybe something you dont want to cover in this PR, but I'd recommend you to try docker buildkit as an image builder, with caching enabled, as described [here](https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching). That will improve performance not just for this stage, but also for the previous ones

.github/workflows/ci.yaml
```diff
@@ -155,6 +151,48 @@ jobs:
...
           docker build -f docker/pmbmpd/Dockerfile -t pmbmpd:_build .
           docker build -f docker/pmtelemetryd/Dockerfile -t pmtelemetryd:_build .
```
dcaba, line 153: unrelated potential improvement: maybe you just want to iterate with a for loop and build all these flavors given a list of daemons, rather than all this almost duplicated `docker build` statements

.github/workflows/ci.yaml
```diff
@@ -155,6 +151,48 @@ jobs:
...
+          echo "Saving images as artifacts..."
+          mkdir -p /tmp/docker/
+          docker save -o /tmp/docker/pmacct_docker_images.tar base:_build pmacctd:_build nfacctd:_build sfacctd:_build uacctd:_build pmbgpd:_build pmbmpd:_build pmtelemetryd:_build
+
+      - name: Export pmacct docker images as an artifact
+        uses: actions/upload-artifact@v2.2.2
```
dcaba, line 159: im an ignorant regarding this, but: is there any chance that two parallel PRs conflict with each other / replace the images we are saving in this artifact storage resource? I want to think that something already adds versioning for you, so one pipeline execution does not reuse the artifacts of another build, but just raising the hand, in case we are missing something

dcaba (approved):
Some feedback shared: most remarkable is that I would not wait until the end of the testing to generate the images, and just block the release stage until the testing is successful; that will reduce the pipeline execution time. But looks a very nice and natural improvement!!

.github/workflows/ci.yaml
msune: Thanks for the pointer. Definitely out of the scope of this PR.

As the article mentions, you need to use an external docker image repository and push it, to get the advantage. I will have a look, as for the builders, specially, that could speed up quite a bit the builds.

For the pmacct containers, I am not sure. Probably the common part, but all these derived containers just change the entry point so building is pretty cheap.

.github/workflows/ci.yaml
msune: https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts

Artifacts are per Github Actions workflow (similarly, say in Gitlab artifacts are per pipeline).

.github/workflows/ci.yaml
msune: Yes, good idea. I will remember for the next CI iteration.

.github/workflows/ci.yaml
msune: I will try to add a commit without this dependency, but I am not sure it will speed up things; you have a number of workers, and these are shared by all "pipelines" (workflows) in github.

Besides, the idea was that - once pmacct has tests - it's not really worth to build some docker containers if tests are not passing, beyond perhaps giving early notice to developers. Thing is, pmacct docker stage is pretty "simple", and it's one of the build combinations already tested in the previous stage, so it's unlikely that it's broken (unless the CI code itself is broken).

I will give it a quick test nevertheless, in case it speeds up things.

msune:
(@paololucente Please don't merge, parallel tests in progress in the same branch - only master works if I want to test pipeline e2e)

msune:
@dcaba I've pushed a commit on top with your suggestion, slightly modified. I made the dockerhub job depend on both the `pmacct-docker` and the `build-and-test`, so that we never publish with `bleeding-edge` tag any build that didn't pass the tests (none, beyond compiling at the moment).

@paololucente ready to merge

PR #518, opened 2021-07-30
pmacct: use correct IP option for jumbo payload packets

olk:
According to RFC 2292 'Advanced Sockets API for IPv6' datagrams
with more than 65535 bytes payload require the jumbo payload option,
e.g the length of the payload ist stored in the hop-by-hop header.
In order to exclude jumbo payload packets we have to check if the
IP6 header payload length is zero and the next header is a
hop-by-hop header.

### Checklist

I have:


paololucente (commented):
Thank you Oliver ( @olk ), makes sense & i will merge this asap. Paolo

PR #522, opened 2021-08-06
Add contextual out-of-bounds checks for maxLength

job:
### Short description

The RTR PDU element for maxLength is `uint8`, therefor it might be possible to transmit incorrect RTR data. `pmacct` should be aware of this situation

### Checklist
- [x] compiled
- [ ] tested this code


paololucente (commented):
Thank you Job for this contribution. Makes sense & I will merge right away. Paolo

PR #529, opened 2021-09-14
Implement aggregate_unknown_etype config key and functionality.

singularsyntax:
Hello Paolo,

Per our discussion in email, I've made changes to support aggregation/counting of unknown EtherType(s) by `pmacctd`.

### Short description
Added `aggregate_unknown_etype` boolean config key, which defaults to false, and which if enabled makes Ethernet frames with unsupported EtherTypes "visible" to the `pmacctd` aggregation engine.

### Checklist
I have:
- [N/A] added the [LICENSE template](https://github.com/pmacct/pmacct/blob/master/LICENSE.template) to new files
- [Y] compiled & tested this code
- [Y] included documentation (including possible behaviour changes)

### DIFF
--- a/src/pkt_handlers.c
+++ b/src/pkt_handlers.c
@@ -1347,6 +1347,7 @@ void counters_handler(struct channels_list_entry *chptr, struct packet_ptrs *ppt
 
   if (pptrs->l3_proto == ETHERTYPE_IP) pdata->pkt_len = ntohs(((struct pm_iphdr *) pptrs->iph_ptr)->ip_len);
   else if (pptrs->l3_proto == ETHERTYPE_IPV6) pdata->pkt_len = ntohs(((struct ip6_hdr *) pptrs->iph_ptr)->ip6_plen)+IP6HdrSz;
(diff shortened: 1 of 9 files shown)


src/pkt_handlers.c
```diff
@@ -1347,6 +1347,7 @@ void counters_handler(struct channels_list_entry *chptr, struct packet_ptrs *ppt
 
   if (pptrs->l3_proto == ETHERTYPE_IP) pdata->pkt_len = ntohs(((struct pm_iphdr *) pptrs->iph_ptr)->ip_len);
   else if (pptrs->l3_proto == ETHERTYPE_IPV6) pdata->pkt_len = ntohs(((struct ip6_hdr *) pptrs->iph_ptr)->ip6_plen)+IP6HdrSz;
+  else if (config.aggregate_unknown_etype) pdata->pkt_len = pptrs->pkthdr->len-ETHER_HDRLEN;
```
singularsyntax, line 1350: I considered setting the unknown Ethernet frame "packet length" to `pptrs->caplen` instead of `pptrs->len` (less the Ethernet header length). I'm not sure which is the best choice here, but I decided on the original packet length since we're throwing away the payload on aggregation anyway, and it's probably more useful to the user to have the actual packet length in the aggregation purge events...

PR #533, opened 2021-09-28
Docker build

rbarazzutti:
### Short description
Refactoring of Dockerfile/build scripts:
- in order to speed up the build (re-order of the operations)
- 3rd party libraries dependencies pointing to specific versions instead of the latest version.
- some minor refactoring

### Checklist

I have:


paololucente (commented):
Hi Raphael ( @rbarazzutti ),

Thanks very much for this much appreciated contribution.

I support clamping the dependencies to specific / tested versions taking on us the task to advance them to latest stable from time to time; this is something that did already bite us in the past, hence SOME of the packages were clamped whereas others were not.

I am also fine with the structuring of the gh_* functions in `deps.sh`. I'd defer to the great Marc ( @msune ) what he thinks about the changes in the Dockerfile. Marc? Comments?

Paolo

msune:
@rbarazzutti First of all thank you for this PR.

On fixing or clamping the version of the libraries... absolutely! I didn't realise some were not checking out a particular tag or commit, sigh. It has been pure luck the `HEAD` of their main branch did not change or break APIs, or introduce regressions. Sort of related but tangential topic, you can read here: https://github.com/pmacct/pmacct/pull/503#issuecomment-842306987 what are my thoughts on how dependecies could be handled in pmacct, avoiding this problem in general.

On the `gh_` functions, seems like a good idea. I will do some minor comments in the code.

On the reordering of the copying of `/usr/local/`, I am not sure what's the rationale there.. My gut feeling would think that this is slower, not faster, because there are two extra RUN statements (`ldconfig` and removing `apt.lists`). I will try it locally see if I understand that change better (cc @dcaba)

ci/deps.sh
```diff
@@ -34,40 +35,72 @@ echo "MAKEFLAGS: ${MAKEFLAGS}"
...
+gh_retrieve_wget(){
+    ORG=$1
+    REPO=$2
+    TAG=$3
+
+    echo "Downloading $ORG $REPO $TAG"
+    
+
+
+    wget $WGET_FLAGS https://github.com/$ORG/$REPO/archive/refs/tags/$TAG.tar.gz && \
```
msune, line 47: For now, there are only references to github repositories (except for AVRO, but you moved it to use the official repo in github). That might not always be the case, and then this function should be rewritten / write another one. I would pass the entire URL.

ci/deps.sh
```diff
@@ -34,40 +35,72 @@ echo "MAKEFLAGS: ${MAKEFLAGS}"
...
-cd rabbitmq-c-0.11.0 ; rm -rf ./.git ; mkdir build ; cd build ; cmake -DCMAKE_INSTALL_LIBDIR=lib .. ; sudo cmake --build . --target install ; cd .. ; cd ..
+gh_build "alanxz" "rabbitmq-c" "v0.11.0" "mkdir build && cd build && cmake -DCMAKE_INSTALL_LIBDIR=lib .. && sudo cmake --build . --target install && cd .." && \
 
-git clone --recursive https://github.com/maxmind/libmaxminddb
-cd libmaxminddb ; rm -rf ./.git ; ./bootstrap ; ./configure ; make ; sudo make install ; cd ..
+GIT=1 gh_build "maxmind" "libmaxminddb" "1.6.0" "./bootstrap && ./configure && make && sudo make install" && \
 
-git clone -b 3.4-stable https://github.com/ntop/nDPI
-cd nDPI ; rm -rf ./.git ; ./autogen.sh ; ./configure ; make ; sudo make install ; sudo ldconfig ; cd ..
+GIT=1 gh_build "ntop" "nDPI" "3.4-stable" "./autogen.sh && ./configure && make && sudo make install && sudo ldconfig" && \
```
msune, line 93: Given that you are building helper functions, but that these are not very large, it might be worth defining `build_git` and `build_wget` instead of passing this env. variable (and in any case I would prefer a parameter than an env. variable in this case)

msune (commented):
Minor comments

ci/deps.sh
rbarazzutti: Hi Marc,
Thanks for the idea, I'll try to have a more generic approach in case we have something that is not hosted on Github.

ci/deps.sh
rbarazzutti: Hi Marc,
Thanks, I'll try to address this at the same time as your comment above.

rbarazzutti:
Thanks so much, Paolo and Marc for your reviews.

Marc, about speed, this change makes the build quicker in case of a subsequent build after a change in the source code of pmacct. 

The following build command of `docker/base/Dockerfile` doesn't need to run again when a change is done in the code source of pmacct.
```
# Runtime deps
RUN apt-get update && \
  apt-get install -y \
    libmariadb3 \
    libnuma1 \
    libcurl4 \
    libpcap0.8 \
    libpq5 \
    libsnappy1v5 \
    libsqlite3-0 \
    libssl1.1 && \
  apt-get -y clean && \
  rm -rf /var/lib/apt/lists/* && \
  ldconfig
```
It saves some build time and some storage space since builds share one more common layer.

And bonus, we are not querying Debian's servers at each build.

In my local CI-CD pipeline, it reduced the build time by 16% (from 376 to 326 secs) and also reduced the size of the changes between two images by 5%. 

Cheers

msune:
@rbarazzutti Thank you for your explanation. I think I understand now the point, and makes sense.

I tried to do some measurements myself, but I believe my results are not relevant, as I guess the variance is higher than this 5%, because the first build with the patch is 11s slower, which doesn't make sense to me... (should be very similar).

In any case, the cached build seems to be consistently a bit faster. Mind though, pmacct's CI will not benefit from this, as the CI itself runs in a docker container, so there is no reusable cache (we would have to use an external artifact to reuse layers across jobs/builds).

ACK from my side on this specific part of the review.

## Details on the measurements.

For the record, I tried doing `docker system prune` to remove all cache layers etc.. before every First build.

### Without patch

Size: 303
First build:
```
Successfully built 84e306b48084

real	7m0.551s
user	0m0.792s
sys	0m0.549s
```
Second build:

```
Successfully built f005664e979b

real	2m16.780s
user	0m0.541s
sys	0m0.329s
```

### With patch

Size: 303
First build:

```
Successfully built 7ec66521c50b

real	7m11.957s
user	0m0.709s
sys	0m0.585s
```

Second build:

```
Successfully built 1f0739a26572

real	2m4.311s
user	0m0.432s
sys	0m0.350s
```

ci/deps.sh
```diff
@@ -34,40 +35,72 @@ echo "MAKEFLAGS: ${MAKEFLAGS}"
...
+    cd .. && \
+    
+    rm -rf $REPO    
+}
+
+
 # Dependencies (not fulfilled by Dockerfile)
-git clone https://github.com/akheron/jansson
-cd jansson ; rm -rf ./.git ; autoreconf -i ; ./configure ; make ; sudo make install ; cd ..
+gh_build "akheron" "jansson" "v2.13.1" "autoreconf -i && ./configure && make && sudo make install" && \
```
dcaba, line 85: Up to the project owners, but I would rethink the concatenation of all script lines with that final `&& \`; any command that breaks that "pipeline" / line we may forget to end with the `&& \` sequence will mean that any error generated from that line or the previous ones will be ignored, so very prone to human errors. I think that, if you just skip that final `&& \`, you already have the behavior you want: any `gh_build` call that returns a non 0 exit code will mean that the script will stop executing and return that exit code thanks to the `-e` flag, while in the future, adding lines will not require to remember this trick

ci/deps.sh
```diff
@@ -34,40 +35,72 @@ echo "MAKEFLAGS: ${MAKEFLAGS}"
...
+    TAG=$3
+
+    echo "Downloading $ORG $REPO $TAG"
+    git clone --branch $TAG --recursive https://www.github.com/$ORG/$REPO
+}
+
+gh_build(){
+    ORG=$1
+    REPO=$2
+    TAG=$3
```
dcaba, line 65: What if the script supported a default value for the tag (like master/main)? That would require reordering the args; again, up to the project owners, but maintaining version pinning requires focus, and the chance of shipping software with (even security) known bugs can be a problem... One approach I have been using in projects where we don't have the time to timely push versions of our deps is "pull latest versions of our deps and rely on continuous testing to identify breaking changes the sooner the better; fallback to specific pins if we detect a breaking change and we cannot immediately adapt our code", but not saying this approach is the right one for this project... a pity we dont have a proper deps resolver that let's us to cleanly implement hybrid models (for A pick the latest; for B pick any 1.x; for C pick 2.1.4).

dcaba:
> e patch is 11s slower

This may be explained by the fact that the new Dockerfile ships an extra layer with the image (as there is one extra RUN command, the final ldconfig; in the original implementation, we were able to concatenate that to the deps installation step, given the old ordering). But I think the proposal makes sense; its ok to prioritize caching of that deps-installation step.

ci/deps.sh
msune: agreed with @dcaba. I think previous pull requests were addressing this, so this would be a regression

msune:
@dcaba 

> But I think the proposal makes sense; its ok to prioritize caching of that deps-installation step.

I hope it's not too late... I think I disagree. Very few people - to my understanding - builds docker images for pmacct. CI does not currently cache layers, all containers are built from scratch.

If layers would be cached somehow (and that requires a separate discussion and it's more work), it's still more important IMO the overall size of the container than the pipeline speed, if I have to choose. It is both speed to download and boot pmacct containers and BW required, if you have to do it frequently on new nodes.

PR #552, opened 2021-11-30
Implementing AVRO Unions to handle null label's values for both BMP & BGP

scuzzilla:
### Summary of the new functionalities

The previously developed functions, utilized to define the pretag-map's label field shaped as MAP, are now refactored to be able to extend the MAP label's field to both BMP & BGP protocols. 
As you can see from the prototypes here below a new "boolean parameter" (int) has been added to be able to toggle the functions support between IPFIX & BMP/BGP:

```C
/* prototypes */
extern void compose_label_avro_schema(avro_schema_t, int);
extern int compose_label_avro_data(char *, avro_value_t, int);
```
the boolean set to FALSE is used to activate the right behavior for IPFIX while TRUE is the right value for BMP/BGP.

the main difference between the two behaviors is the way the label's null values are handled:
- for what concerning BMP/BGP, "avro unions" are used to define the map's default value to null (This is necessary to cover the BMP/BGP messages with null label information associated).
- for what concerning IPFIX, the default value is not necessary since the label field is either with or without value but *never* set to null.

### Checklist

I have:

### DIFF
--- a/src/plugin_cmn_avro.c
+++ b/src/plugin_cmn_avro.c
@@ -36,7 +36,7 @@
 #ifdef WITH_AVRO
 /* global variables */
 avro_schema_t p_avro_acct_schema, p_avro_acct_init_schema, p_avro_acct_close_schema;
-avro_schema_t sc_type_map, sc_type_string;
+avro_schema_t sc_type_map, sc_type_string, sc_type_union;
 
 /* functions */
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thank you Salvatore ( @scuzzilla ) for this important contribution in order to bring Apache Avro back on par with JSON functionality for BGP and BMP now that pre_tag_map support has been introduced. I reviewed, all looks good & i will merge imminently. Paolo

PR #560, opened 2022-01-13
MySQLv8 problem with DATETIME types

drushadrusha:
### Short description

When I'm trying to run sfacctd with mysql plugin it throw errors like:
`Incorrect datetime value: '1642073401' for column 'stamp_updated' at row 1`

MySQL datetime is not supposed to store unix timestamps.
Changed columns 'stamp_inserted' and 'stamp_updated' to bigint and now it works.

In other way, we can pass UNIX_TIMESTAMP() function in INSERT's and keep datetime types, but in my opinion timestamps more easy to deal with than mysql date types.
### Checklist

I have:

### DIFF
--- a/sql/pmacct-create-db_v8.mysql
+++ b/sql/pmacct-create-db_v8.mysql
@@ -22,7 +22,7 @@ create table acct_v8 (
     packets INT UNSIGNED NOT NULL,
     bytes BIGINT UNSIGNED NOT NULL,
     flows INT UNSIGNED NOT NULL,
-    stamp_inserted DATETIME NOT NULL,
-    stamp_updated DATETIME,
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Dear @drushadrusha ,

Thank you for your contribution. I guess you had to set `timestamps_since_epoch` to true for this. Not being the default setting, 1) default schemas are shipped as DATETIME type and 2) timestamps_since_epoch documentation indicates that the type should be defined as BIGINT. So i will close this issue and not merge it.

True is that maybe this type change could be given more relevance also in other pieces of documentation, ie. README.mysql or README.timestamp in the sql/ tree. I will work on that.

Paolo

PR #562, opened 2022-01-18
fix the memleak created by label(MAP) + AVRO + BxP (msglog) ...

scuzzilla:
### Changelog

- The function "avro_generic_value_new()" has been deleted from each single sub-function related to label(MAP) | TCP-FLAGs | ... : the AVRO's interfaces & associated values (including their decref) are handled by the upper-level functions. (memory leak main issue).

- The function in charge to serialize the LABEL(MAP) data to KAFKA is now divided into two separated functions: the first one for IPFIX while the second one for BxP (BGP/BMP)

- The exact same approach as above is now adopted for the schema generation, there are now two separated functions: one for IPFIX & one for BxP.

-  The function in charge of generating the linked-list storing the label(MAP) information has been refactored replacing strdup() with strcpy().

### Checklist

I have:

### DIFF
--- a/src/bgp/bgp_logdump.c
+++ b/src/bgp/bgp_logdump.c
@@ -37,6 +37,7 @@
 #include "plugin_cmn_avro.h"
 #endif
 
+
 /* functions */
 int bgp_peer_log_msg(struct bgp_node *route, struct bgp_info *ri, afi_t afi, safi_t safi,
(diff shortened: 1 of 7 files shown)


paololucente (commented):
Thank you Salvatore ( @scuzzilla ) for this precious code consolidation. Will merge asap. Paolo

PR #597, opened 2022-04-15
build: fix git safe directory errors

msune:
### Short description

As a result of patches introduced in `git` to protect against
the defects CVE-2022-24765, CVE-2022-24767, pipelines started to
fail due to the new `git` versions being rolled out, with the error
`Error: fatal: unsafe repository`.

This commit fixes this by setting the appropriate exceptions for
the repository itself and for the git submodules

### Checklist

I have:


paololucente (commented):
Thank you Marc ( @msune ) for jumping on and resolving this tricky issue so quickly. Merging. Paolo

PR #619, opened 2022-07-02
Don't set the AS Path for receiving packets in flows in sflow data

WRMSRwasTaken:
### Short description
This is a (very late, sorry about that, things didn't turn out quite as expected) followup to https://www.mail-archive.com/pmacct-discussion@pmacct.net/msg04121.html where my sflow collector [NetMeta](https://github.com/monogon-dev/NetMeta) ([goflow](https://github.com/cloudflare/goflow) based) fails to process the ASN information found in sflow packets generated by pmacct.

### Checklist

I have:

### What am I doing here?
Looking at the goflow2 [source code](https://github.com/netsampler/goflow2/blob/7d4c41a161d707922b30bfbb7752cd35d2f4b4eb/producer/producer_sf.go#L300) it can be seen, that it is populating the source AS field based on the length of the AS Path in the sflow extended gateway data. If it is greater than zero, the source AS field will be set based on the "AS Router" field.
pmacct always sets the AS path to one with the destination AS being the only element even if it's zero, as seen [here](https://github.com/pmacct/pmacct/blob/7e5b36564ed2b7a43f561df4620f2b81f3915546/src/sfprobe_plugin/sfprobe_plugin.c#L535).
However, the "AS Router" field is always zero, at least for me:
grafik
... this will always have the source AS field set to zero due to goflow (1 & 2) always setting the source AS based on the router AS.
Looking at the memory plugin, pmacct seems to account this correctly (not having an AS path for receiving flows):
grafik

I've added a check to see if the destination AS is not zero in order to put the destination AS into the AS path for sflow packets, which effectively causes the AS path only getting populated for sending flows.

I was looking at putting the full AS path into the sflow data (not just the destination AS as the only element), but the `pkt_payload` struct doesn't seem to carry the AS path itself.

This is just a draft in order to test things further and to see if this is the only needed change for my collector to start processing ASN information. A quick test run with running raw goflow2 and printing the decoded sflow data to console yielded successful results.

### DIFF
--- a/src/sfprobe_plugin/sfprobe_plugin.c
+++ b/src/sfprobe_plugin/sfprobe_plugin.c
@@ -529,11 +529,18 @@ static void readPacket(SflSp *sp, struct pkt_payload *hdr, const unsigned char *
 	gatewayHdrElem.tag = SFLFLOW_EX_GATEWAY;
 	// gatewayHdrElem.flowType.gateway.src_as = htonl(hdr->src_ip.address.ipv4.s_addr);
 	gatewayHdrElem.flowType.gateway.src_as = hdr->src_as;
-	gatewayHdrElem.flowType.gateway.dst_as_path_segments = 1;
-	gatewayHdrElem.flowType.gateway.dst_as_path = &as_path_segment;
-	as_path_segment.type = SFLEXTENDED_AS_SET;
-	as_path_segment.length = 1;
-	as_path_segment.as.set = &hdr->dst_as;
+
+        if (hdr->dst_as == 0) {
+          gatewayHdrElem.flowType.gateway.dst_as_path_segments = 0;
+        }
+        else {
+          gatewayHdrElem.flowType.gateway.dst_as_path_segments = 1;
+          gatewayHdrElem.flowType.gateway.dst_as_path = &as_path_segment;
+          as_path_segment.type = SFLEXTENDED_AS_SET;
+          as_path_segment.length = 1;
+          as_path_segment.as.set = &hdr->dst_as;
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Hi Marcel ( @WRMSRwasTaken ),

This patch does make sense, thanks very much for submitting it. Let me know when you believe it is good to merge, for me it is.

I read that you are marking it as a draft while looking whether there are other things to commit but, unless strictly related to this, since opening PRs is like printing new money .. it comes for free, i recommend to not conflate patches in the same PR. 

Paolo

WRMSRwasTaken:
Hello Paolo,

the code in this PR is staying as it is, if I think there is something other to change I'd go ahead and open a new PR, yes the message was a bit unclear regarding this.

I am hoping to get some more input from people I've asked to verify this new behavior with the added code.
At least from what I've tested with goflow2 so far, the new behavior makes sense for me, so I am planning to un-draft this later today or tomorrow.

 - Marcel

paololucente:
Perfect Marcel, did you by any chance already tested the patch against `sfacctd` as well? If not, i will do it. Paolo

paololucente:
I have just tested it against `sfacctd` and i can confirm that it works. Not only works but it's also very beneficial. At least in JSON encoding, behavior of the collector changes from  `"as_path": "0"` (not nice!) to `"as_path": ""` (nice!) with your patch.

WRMSRwasTaken:
I checked some more things with goflow2 (like generating sflow samples from transiting data to downstreams) and couldn't find any issues with the new behavior.

PR #620, opened 2022-07-03
Fix multistage docker build (bleeding-edge). Add smoke test

msune:
### Short description

Bleeding edge containers have been broken since ndpi version was bumped. This was reported by Lisa Ensman (thanks) in te docker-doctors mailing list.

Unfortunately, bleeding-edge containers didn't have testing as stable do (during the release process). 

This PR does two things:

a) It adds a smoke test to make sure all containers start properly, using docker compose. This can be the taken as the basis for other tests.

b) Fixes the build.

Details for b) here:

```
    ci|docker: fix docker multi-stage build
    
    Two problems are addressed with this commit:
    
    * After bumping libndpi to 4.2, default installation path is
      no longer /usr/local, hence the library is not copied from the
      build container to the final base container. Fixed by forcing
      autotools and cmake pkgs to use /usr/local
    * In the final base container libjson-c andlibnetfilter-log1 were
      missing. Added to the final base container.
    
    Signed-off-by: Marc Sune <marcdevel@gmail.com>
```

### Checklist

I have:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -166,6 +166,18 @@ jobs:
           mkdir -p /tmp/docker/
           docker save -o /tmp/docker/pmacct_docker_images.tar base:_build pmacctd:_build nfacctd:_build sfacctd:_build uacctd:_build pmbgpd:_build pmbmpd:_build pmtelemetryd:_build
 
+      - name: Docker (compose) smoke test
+        run: |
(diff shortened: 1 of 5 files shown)


paololucente (commented):
Thank you Marc ( @msune ), merging imminently. Thanks also for the outstanding work having taken & always keeping both CI and Docker things in pmacct to a fantastic level. Paolo

PR #625, opened 2022-07-19
plugin_cmn_avro: change the logic to generate schema names

ustorbeck:
This request changes the logic to generate schema names:

- use topic plus record name as the schema name
- use underscore as separator within the record name
- stop to add a "-value" suffix

I have compiled & tested this code.


paololucente (commented):
Hi Uwe ( @ustorbeck ),

Thank you very much for this contribution!

While these changes are not backward compatible, given the scope of the feature, they can be handled via a note in the `UPGRADE` document - i can take care of that. Before proceeding to a merge, I was hopefully looking for two pieces of information from you in order to make myself a better idea on the context of these proposed changes:

* Why changing the `-` separator to `_`? Is it to have two different separators? Which, in general terms, it's always a good idea?
* Can you comment on why removing the `-value` string? A-la, why was it needed, why is it not needed anymore?

Paolo

PR #632, opened 2022-08-08
Avoid to cut off log messages

ustorbeck:
When log messages are longer than our message buffer the message gets cut off. As the trailing newline also gets cut off the message will be concatenated with the following message which makes the log hard to read. Avoid to cut off the messages by not using an intermediate buffer.

I've tested the changes on top of commit ce0d5557 and to a lesser extend also on top of 57e5b892 and fa0139fb.


paololucente (commented):
Hi Uwe ( @ustorbeck ), thanks very much for this contribution. I see its usefulness and code changes do make sense to me. I will do some testing at my end as well and merge asap. Paolo

PR #633, opened 2022-08-08
Reduce the number of requests being sent to the schema registry

ustorbeck:
Move the schema registration from the writer to the plugin process to register the schemas only once at the start of the Kafka plugin process and not on every start of a writer process.

I've tested the changes with nfacctd on top of commit ce0d5557 and to a lesser extend also on top of 57e5b892 and fa0139fb.

### DIFF
--- a/src/kafka_plugin.c
+++ b/src/kafka_plugin.c
@@ -225,6 +225,14 @@ void kafka_plugin(int pipe_fd, struct configuration *cfgptr, void *ptr)
       compose_avro_schema_registry_name_2(config.sql_table, FALSE,
                                           p_avro_acct_close_schema, "acct", "close",
                                           config.kafka_avro_schema_registry);
+
+    if (!kafkap_kafka_host.sd_schema[AVRO_ACCT_DATA_SID] ||
+        !kafkap_kafka_host.sd_schema[AVRO_ACCT_INIT_SID] ||
+        !kafkap_kafka_host.sd_schema[AVRO_ACCT_CLOSE_SID]) {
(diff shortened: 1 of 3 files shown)


src/kafka_plugin.c
```diff
@@ -194,6 +194,48 @@ void kafka_plugin(int pipe_fd, struct configuration *cfgptr, void *ptr)
...
+
+#if defined(WITH_AVRO) && defined(WITH_SERDES)
+  if (config.kafka_avro_schema_registry) {
+    if (strchr(config.sql_table, '$')) {
+      Log(LOG_ERR, "ERROR ( %s/%s ): dynamic 'kafka_topic' is not compatible with 'avro_schema_registry'. Exiting.\n",
+          config.name, config.type);
+      exit_gracefully(1);
+    }
+
+    p_kafka_init_host_struct(&kafkap_kafka_host);
```
paololucente, line 214: Hi Uwe ( @ustorbeck ),

Thanks for this very useful contribution.

The whole patch does makes sense to me; the only thing that leaves me puzzled is using p_kafka_init_host_struct() inside the if section: should the structure not be initialized regardless? Exactly like p_kafka_init_host_conf() is unconditionally executed in kafka_cache_purge()?

Paolo

ustorbeck:
Hi Paolo

Good catch. Thank you!
I've updated the pull request.

Uwe

paololucente (commented):
Thank you Uwe, it now looks perfect to me. I will do some testing at my end and - all behaving well as it looks like - i'll merge asap. Paolo

PR #635, opened 2022-08-24
Do not set partition to RD_KAFKA_PARTITION_UA if static kafka_partition was configured

rodecker:
### Short description
A check for `config.kafka_partition` was missing. This lead to the plugin always using the default partitioner instead of sending to the configured fixed partition.

### Checklist

I have:
- [N/A] added the [LICENSE template](https://github.com/pmacct/pmacct/blob/master/LICENSE.template) to new files
- [V] compiled & tested this code
- [N/A] included documentation (including possible behaviour changes)

### DIFF
--- a/src/kafka_plugin.c
+++ b/src/kafka_plugin.c
@@ -462,7 +462,7 @@ void kafka_cache_purge(struct chained_cache *queue[], int index, int safe_action
     exit_gracefully(1);
   }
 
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thank you Martin ( @rodecker ), that makes total sense. Will merge imminently. Paolo

Issue #638, opened 2022-09-14
uacctd segfault on 1.7.6-2


levin1:
When i use ICMP traffic there is a segfault in function find_flow line `memcpy(&my_tlh, pptrs->tlh_ptr, MyTCPHdrSz);` not sure how to fix this.

ikle:
Proposed quick fix: https://github.com/pmacct/pmacct/pull/639

paololucente:
Thanks @levin1 for raising this issue and thanks Alexei ( @ikle ) for proposing the quick fix and for the subsequent conversation about it, it makes sense. Just merged #639 and going to close this issue. Paolo

PR #639, opened 2022-09-14
src/nl: do not try to handle TCP flow for non-TCP, prevent DDoS

ikle:
The uacctd daemon crashes in find_flow when copying tlh_ptr upon receiving an ICMP Echo Request packet (as a result of ping). Thus, anyone who can send an ICMP Echo Request to our system can bring down the traffic accounting system: this is DDoS. The suggested solution is a quick fix that fixes the issue in our case, no detailed study has been done.

### DIFF
--- a/src/nl.c
+++ b/src/nl.c
@@ -309,9 +309,9 @@ int ip_handler(register struct packet_ptrs *pptrs)
         if (((struct pm_tcphdr *)pptrs->tlh_ptr)->th_flags & TH_FIN) pptrs->tcp_flags |= TH_FIN;
         if (((struct pm_tcphdr *)pptrs->tlh_ptr)->th_flags & TH_RST) pptrs->tcp_flags |= TH_RST;
         if (((struct pm_tcphdr *)pptrs->tlh_ptr)->th_flags & TH_ACK && pptrs->tcp_flags) pptrs->tcp_flags |= TH_ACK;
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Hi @ikle ,

Thank you for raising the issua and proposing this fix.

Did you apply it against 1.7.6 release? I ask you because supposedly the issue you are running into in #638 is already fixed in current code with https://github.com/pmacct/pmacct/commit/73af9545ea33cd87846306f648f634063ac41765 . I was hence wondering if this is a problem that that fix does not catch or you are running some old(er) code and can give a try to upgrade it to current.

Paolo

ikle:
> Did you apply it against 1.7.6 release?

Yes, more precisely on the version in Debian 11 (1.7.6-2). Then issue #368 can be closed as a duplicate.

Still, does it make sense to call ip_flow_handler for l4_proto ≠ IPPROTO_TCP?

ikle:
> Still, does it make sense to call ip_flow_handler for l4_proto ≠ IPPROTO_TCP?

For example, in networks with active use of VPN (UDP with OpenVPN, AH, ESP with IPSec) and other tunnels (IPIP, GRE, MPLS), with this change, ip_flow_handler will no longer be called, which will reduce (slightly) the load on the CPU.

Issue #640, opened 2022-09-29
netflow v9: timestamp_end sometimes before timestamp_start


jwestfall69:
**Description**
I've noticed that we sometimes get a timestamp_end that is before timestamp_start, which means a negative flow duration.  The difference is usually around -4290000 when this happens.  For example

```
{"event_type": "purge", ... , "timestamp_start": "1664491696.000000", "timestamp_end": "1660196759.000000", "packets": 2, "bytes": 824}
```

These flows are coming from juniper mx routers.  

**Version**
This was originally seen on an older version but also happens with master..

```
NetFlow Accounting Daemon, nfacctd 1.7.8-git [RELEASE]

Arguments:
 '--enable-jansson' '--enable-l2' '--enable-traffic-bins' '--enable-bgp-bins' '--enable-bmp-bins' '--enable-st-bins'

Libs:
cdada 0.4.0
libpcap version 1.10.0 (with TPACKET_V3)
jansson 2.13.1

Plugins:
memory
print
nfprobe
sfprobe
tee

System:
Linux 5.10.0-11-amd64 #1 SMP Debian 5.10.92-2 (2022-02-28) x86_64

Compiler:
gcc 10.2.1
```

**Debug info**
I added some debug output in nfacctd to print out the raw data related to timestamp_end as its being calculated to track down if its the router or nfacctd causing the problem.  It results in stuff like

```
fstime: 66888064, unix_secs: 1664488348, sysuptime: 66892000, result: 1664488345
fstime: 66884992, unix_secs: 1664488348, sysuptime: 66892000, result: 1664488341
fstime: 66892160, unix_secs: 1664488348, sysuptime: 66892000, result: 1660193381 <----
fstime: 66879872, unix_secs: 1664488348, sysuptime: 66892000, result: 1664488336
fstime: 66881408, unix_secs: 1664488348, sysuptime: 66892000, result: 1664488338
```

fstime and sysuptime are both msec values.  It appears the router is providing its sysuptime in second resolution, but the fstime with msec resolution.  This is resulting in cases where fstime > sysuptime and causing a bad calculated value.

```
      pdata->time_end.tv_sec = ntohl(((struct struct_header_v9 *) pptrs->f_header)->unix_secs)-
        ((ntohl(((struct struct_header_v9 *) pptrs->f_header)->SysUptime)-ntohl(fstime))/1000);
```
which is basically

```
timestamp_end = unix_secs - (sysuptime - fstime)/1000
```

[...]

PR #641, opened 2022-10-11
nf9: adjust timestamp calc to avoid uint32_t underflow

jwestfall69:
### Short description
Tell the compiler to treat the result of (SysUptime - fstime) as signed to avoid an uint32_t underflow in the case where fstime > SysUptime.

This works around an issue (#640) seen on juniper MX routers that was causing timestamp_end to be ~50 days before timestamp_start.

### Checklist

I have:


paololucente (commented):
Thank you Jim ( @jwestfall69 ) for this contribution, as we discussed in #640 this makes sense. I will run some regression testing against it and, all going fine (i don't expect it does not) i will merge asap. Paolo

PR #645, opened 2022-11-07
build: speed up incremental builds after git changes

ustorbeck:
Avoid to rebuild too many object files after git changes. Include the generated header file pmacct-build.h only in source files where the build information actually is used.

### DIFF
--- a/examples/lg/pmbgp.c
+++ b/examples/lg/pmbgp.c
@@ -23,6 +23,7 @@
 
 /* includes */
 #include "pmacct.h"
+#include "pmacct-build.h"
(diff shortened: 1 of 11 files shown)


paololucente (commented):
Thank you Uwe ( @ustorbeck ) for this contribution. I see the value of the contribution, i did occasionally stumble upon too much rebuilding for a one-liner change myself - i bless it myself.

Marc ( @msune ), question to you since this is a tangent on building: do you also bless it?

Paolo

msune (approved):

PR #665, opened 2023-02-09
kafka: log the actual reason for file access errors

ustorbeck:
The error message "file not found" is misleading when the process for example runs out of file descriptors.


ustorbeck:
The failed check seems to be an internal problem and not related to this PR. If not, please ping me.
Is there a way to rerun a failed test without pushing new content to a PR?

paololucente (commented):
Thank you Uwe ( @ustorbeck ) for this. Re-running CI, i also believe like you that it was due to other causes. Will let it clear out and merge. Paolo

PR #667, opened 2023-02-21
avro plugin: fix registry match for tunnel tcp flags

ustorbeck:
The primitive for tunnel tcp flags is matched against the wrong registry in the avro plugin. Fix that matching.


paololucente (commented):
Thanks Uwe ( @ustorbeck ) for fixing this clear oversight. Merging. Paolo

PR #668, opened 2023-02-27
BGP: allow MP nexthop length of 48 bytes.

FRIDM636:
**Extend MP nexthop length to support 48 bytes to fit the BGP VPNv6 nexthop length.** , an example is illustrated in the below figure: 

mpnexthop48


paololucente (commented):
Thank you @FRIDM636 for this contribution, makes total sense. Merging.

Paolo

PR #672, opened 2023-03-10
add support for IPFIX v10 MPLS_RD options

rodonile:
### Short description
With IPFIX v10 the RD option packets are not SYSTEM-scoped anymore, but instead on ingress/egress_VRFID. This PR extends the option handlers to support those option packets by:

- setting srcID/observationDomainID scoped xflow table as default
- only fallback to the global xflow table if no mapping is found

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ) for this contribution. I have reviewed and it reads well. I am going to test shortly and merge. Paolo

PR #673, opened 2023-03-10
add support for NF9_EXPORTER_IPV[46]_ADDRESS field in the data packets

rodonile:
### Short description
Currently the NF9_EXPORTER_IPV4_ADDRESS and NF9_EXPORTER_IPV6_ADDRESS fields are only supported by nfacctd when they are sent within the option packets. With this PR, support for these fields in the IPFIX data packets is added. 

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ) for this contribution. Although it is "interesting" to send the exporter IP as part of every single flow, i guess some vendor must be implementing this :-) The patch makes sense & going to merge imminently. Paolo

PR #675, opened 2023-03-13
pmtelemetryd: fix a crash when no dump method is configured

ustorbeck:
When no backend dump method is configured (and telemetry_dump_time_slots is not set) pmtelemetryd crashes with SIGFPE.


paololucente (commented):
Hi Uwe ( @ustorbeck ), thanks very much for catching this. Merging. The exact same would need to be done for BGP and BMP. Would you like to have the honor or shall i? Paolo

PR #676, opened 2023-03-15
fix a crash for bgp and bmp when no dump method is configured, fix the dump interval for pmtelemetyd when using time slots

ustorbeck:
Avoid the division by zero also for bgp and bmp when no dump method is configured.
Fix the dump interval calculation for pmtelemetryd when time slots are used.


paololucente (commented):
Thank you Uwe! ( @ustorbeck ). merging.

PR #677, opened 2023-03-24
BGP/BMP data structures, enriching the hash function with Route Distinguisher (RD)

scuzzilla:
### Summary 

Within the context of BGP/BMP data structures,  the main aim of this PR is to enrich the hash function with a new hash parameter, the Route Distinguisher (RD), in order to minimize the number of collisions. It is crucial to have an efficient hash-table in order to perform data correlation between BGP/BMP & IPFIX(\*).

#### pmacct offers mainly two configuration options to pre-allocate the number of hash-table's buckets
```TEXT
1. bgp_table_per_peer_buckets | bmp_table_per_peer_buckets [1 - 128]  [DEFAULT: 1]
2. bgp_table_peer_buckets     | bmp_table_peer_buckets     [1 - 1000] [DEFAULT: 13]
```

- The buckets configured by the first parameter are hierarchically above the buckets defined by the second one. "useful when much BGP information per peer is received, ie. in case of BGP ADD-PATHs".
- The buckets configured by the second parameter are giving direct access to the RD information used to perform data correlation with IPFIX.

#### To optimize the hash table, load factor < 0.75
1. we can tune the amount of buckets defined by bgp_table_peer_buckets. 
```TEXT
[Load factor (λ) = Number of stored elements (n) / Number of available slots (N)]
```
2. Adding the RD to the hash function contributes to minimizing the amount of collisions (a better data distribution over the hash table is achieved).

By focusing on optimizing the load factor, we can ensure that the hash table operates efficiently, resulting in better overall performance.

#### During functional testing 
I experimented with various combinations of the above configuration parameters while monitoring the CPU level:

```TEXT
| bgp_table_peer_buckets | bgp_table_per_peer_buckets | %CPU    |
+------------------------+----------------------------+---------+
| DEFAULT                | DEFAULT                    | ~25-30% |
+------------------------+----------------------------+---------+
| DEFAULT                |  128                       | ~35-40% |
+------------------------+----------------------------+---------+
| 1000                   | DEFAULT                    | ~10-15% |
+------------------------+----------------------------+---------+
| 1000                   | 128                        | ~25-30% |
+------------------------+----------------------------+---------+
```

#### Additional considerations
1. Increasing the amount of bgp_table_per_peer_buckets is CPU inefficient most probably due to linear iteration over all buckets.
2. Increasing the amount of bgp_table_peer_buckets (keeping in mind the load factor formula) increases the hash table efficiency with the price of more memory consumption.

---

(*) BGP/BMP Data correlation:
```TEXT
                +------------------+ 
      IPFIX --->| pmacct & flow2rd |---> IPFIX + rd  (Example: peer_ip_src, src_ip, dst_ip, (rd))
                +------------------+   

                +----------------------------+
 IPFIX + rd --->|                            |
                | pmacct + correlation func .|---> IPFIX + rd + BGP communities (Example: peer_ip_src, src_ip, dst_ip, (rd), (comms, ecomms, ...))
    BGP/BMP --->|                            |
                +----------------------------+
```

#### Behavior:
1. per IPFIX record match peer_ip_src against flow2rd ---> retrieve the associated rd (if no match rd set-by-default to 0:0:0)
2. enrich the IPFIX record with the associated rd
3. per IPFIX+rd record match:
- peer_ip_src to identify the BGP-RIB belonging to a specific router
- rd against the BGP-RIB's RD field to select the right VRF
- src_ip/dst_ip against the BGP-RIB's Network field to finally retrieve the associated communities
4. enrich the IPFIX+rd record with the associated communities

### Checklist

I have:

### DIFF
--- a/src/bgp/bgp.c
+++ b/src/bgp/bgp.c
@@ -54,7 +54,7 @@ char *std_comm_patterns_to_asn[MAX_BGP_COMM_PATTERNS];
 char *lrg_comm_patterns_to_asn[MAX_BGP_COMM_PATTERNS];
 struct bgp_comm_range peer_src_as_ifrange; 
 struct bgp_comm_range peer_src_as_asrange; 
-u_int32_t (*bgp_route_info_modulo)(struct bgp_peer *, path_id_t *, int);
+u_int32_t (*bgp_route_info_modulo)(struct bgp_peer *, rd_t *, path_id_t *, int);
 struct bgp_rt_structs inter_domain_routing_dbs[FUNC_TYPE_MAX], *bgp_routing_db;
 struct bgp_misc_structs inter_domain_misc_dbs[FUNC_TYPE_MAX], *bgp_misc_db;
 bgp_tag_t bgp_logdump_tag;
--- a/src/bgp/bgp.h
+++ b/src/bgp/bgp.h
@@ -292,7 +292,7 @@ struct bgp_misc_structs {
   int table_per_peer_buckets;
   int table_attr_hash_buckets;
   int table_per_peer_hash;
-  u_int32_t (*route_info_modulo)(struct bgp_peer *, path_id_t *, int);
+  u_int32_t (*route_info_modulo)(struct bgp_peer *, rd_t *, path_id_t *, int);
   struct bgp_peer *(*bgp_lookup_find_peer)(struct sockaddr *, struct xflow_status_entry *, u_int16_t, int);
   int (*bgp_lookup_node_match_cmp)(struct bgp_info *, struct node_match_cmp_term2 *);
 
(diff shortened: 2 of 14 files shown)


paololucente (commented):
Thank you Salvaore ( @scuzzilla ) for this very relevant and important contribution. I have reviewed the code and it makes total sense for environments with L3 MPLS VPNs; i will perform some testing at my end and merge. Paolo

paololucente:
Hi @doup123 ,

No extra configuration is needed, if a RD is present it is taken into consideration for the hashing. Indeed, in turn, to have the RD populated some extra information is needed. If collecting IPFIX, it is possible that the RD is supplied as part of the flow itself or the Options. Otherwise, if a VRF ID is supplied, or in other cases (ie. map ingress interface to a RD), the `flow_to_rd_map` map can be used (you can read more in CONFIG-KEYS [ https://github.com/pmacct/pmacct/blob/6362a2c9e941e8745631d109ecbca5ccf18547aa/CONFIG-KEYS#L2889 ]).

Paolo

PR #681, opened 2023-04-17
netflow: support forwarding status field with size of 2 or 4 bytes

ustorbeck:
This change extends the handling of the forwarding status field when it is filled up with leading zeros to a size of 2 or 4 bytes.


paololucente (commented):
Thank you Uwe ( @ustorbeck ). The proposal makes sense & merging imminently. Paolo

PR #682, opened 2023-04-21
Support correlation with BGP/BMP when RD is in IPFIX/NFv9 options

rodonile:
### Short description
When the RD is sent via IPFIX options, currently nfacctd is not considering it for correlation with BGP/BMP. This PR adds a function that is executed before the bgp-table lookup which loads the RD in the apposite fields to enable correlation this scenario.

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ) for having spotted and fixed this problem. I will review the code and merge. Paolo

PR #683, opened 2023-05-05
RD from options: check also other VRF when RD=0:0:0

rodonile:
### Short description
The current logic for gathering the RD from the option data relies on IE61 flowDirection field, which can be ingress or egress depending on the sampling configuration. With this logic, only one direction of vpnv4/6 flows (the flows coming from the customer facing interface of the PE node) is enriched with the relative RD, while the other direction (flows decapsulated from the MPLS or SR/SRv6 network) is enriched with RD=0:0:0.

This PR adds logic to check option data for the other VRF in cases where we obtain RD=0:0:0, and this makes sure that also the other flow direction is enriched with the actual vpnv4/6 RD.

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ), this increased logic makes sense to pick and associate the right information. I'll review and test briefly at my end and merge. Paolo

PR #685, opened 2023-05-16
Adding support for Extended Next Hop Encoding BGP capability

rodonile:
### Short description
As specified by RFC8950, routers cannot advertise IPv4 or VPNV4 routes with an IPv6 next hop without an extra capability advertisement (Extended Next Hop Encoding). This is relevant for example for VPNV4 routes in an SRv6 core. 

This PR adds support to nfacctd so that it can reply to the BGP OPEN message with this capability, and the router can advertise these routes. It also adds a small bugfix to allow the bgp_nexthop primitive to be populated correctly when AFI differs from next-hop AFI. 

### Checklist

I have:


paololucente (commented):
Thank you as always Leonardo ( @rodonile ) for this contribution. It seems straightforward and does make sense. Let me perform a bit of testing at my end before merging, hopefully already by end of the week. Paolo

PR #693, opened 2023-06-13
remove password from log output

mabezi:
### Short description
It removes passwords from the log.

I mean, why should a password be in plain text in the log?
This is a disaster from a safety perspective.

### Checklist
I have:


paololucente (commented):
Thanks @mabezi , makes total sense. Merging. Paolo

PR #698, opened 2023-06-26
Fix rd selection precedence

rodonile:
### Short description
This PR fixes an issue related to RD preloading that was preventing correlation from IPFIX options to succeed when flow_to_rd map is also enabled in the config.

I also did some cleanup: since we anyway need to preload the RD in nfacctd.c, I made sure that the RD is already loaded with the correct RD_ORIGIN and removed all the duplicated logic in pkt_handlers.c.

The RD precedence is still flow_to_rd.map > RD in IPFIX DATA message > RD in IPFIX OPTION DATA message

Have a look and tell me if it works for you or you want to do something differently :)

Ciao,
Leonardo

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ) for the fix and the reorg of the code. The fix makes sense, I will review, test and come back and/or merge. Paolo

PR #703, opened 2023-07-04
Enhance support for MPLS VPN RD in bgp table hash function

rodonile:
### Short description

This PR enhances the changes made in PR #677 for enabling BGP table hash calculations using the MPLS VPN RD field.

**The changes performed are:**

1. Creating a separate hash function which uses the mpls_vpn_rd instead of path_id as argument and adding _mpls_vpn_rd_ as a parameter for the _bgp_table_per_peer_hash_ config key or _bmp_table_per_peer_hash_ respectively. 

2. Specifically for the bmp_lookup functions, the RD is not present in the bgp_attr_extra structs like it is the case for BGP. Within the BMP lookup functions, the RD for the hash calculations is now gathered from the bgp_msg_extra_data struct.

3. For both BGP and BMP, the modulo function that uses the RD has been slightly adjusted to restore the peer_bucket logic. This fixes the delete capability.

4. Finally, within the bgp_node_match() function (in bgp_table), when using _mpls_vpn_rd_ as hash the modulo can already be calculated precisely using the RD from the IPFIX/NFv9 data. This approach has the advantage that during bgp_table lookup for correlation with IPFIX/NFv9 data, it is no longer required to iterate through all the per_peer_buckets (for a given peer) to find the bucket with the correct NLRI. The loop iterations when using _mpls_vpn_rd_ as _bmp/bgp_table_per_peer_hash_ argument have been limited to 1.

**Example configuration for nfacctd:**
bgp_table_peer_buckets: 15
bgp_table_per_peer_buckets: 10
bgp_table_per_peer_hash: mpls_vpn_rd 

With the above configuration, the BGP information for a specific peer is split within 10 per_peer_buckets and the RD is used as a randomizer parameter to select the bucket.

Cheers,
Leonardo

### Checklist

I have:


rodonile:
Here a small example of bucketing with some BGP updates:

image

paololucente (commented):
Thank you Leonardo ( @rodonile ), useful contribution as always. It seems all in order: i will review more in depth, do some brief testing at my end and merge. Paolo

PR #704, opened 2023-07-13
Change srv6 sid list from suffix to prefix

rodonile:
### Short description
With this PR, the SRv6 sid list encoding is modified so that the indexes are prefixed to the SID, such as:

0-cafe:5::100
1-cafe:2::10e

In the current implementation the indexes are suffixed. This change aims to uniform behavior with the mpls_label_stack encoding, where labels are prefixed with the index.

Ciao,
Leonardo

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ), to ack all. I will briefly review and merge. Paolo

PR #706, opened 2023-07-14
PT_map_label_handler function: replacing 'strtok' (not-reentrant) with 'strtok_r' (reentrant)

scuzzilla:
### Summary 

strtok is not reentrant. This means that if strtok is called from within a function that is itself called while strtok is still splitting a string (for instance, through a signal handler or recursive call), it will start from scratch and lose its place in the original string. 

strtok_r is reentrant because it keeps the state in the user-supplied saveptr argument, allowing you to have multiple tokenizations in progress at the same time in the same thread.

### Checklist

I have:


paololucente (commented):
Thank you Salvatore ( @scuzzilla ) for catching this. Merging & soon onto replacing the handful other strtok() instances with strtok_r(). Paolo

PR #710, opened 2023-07-31
Preload inner/customer layer ip_src/ip_dst for correlation with BGP/BMP [SRv6]

rodonile:
### Short description
In order to perform correlation of IPFIX/NFv9 information with BGP/BMP we need RD + source and destination IP addresses. With an SRv6 underlay we need the inner/customer layer IP addresses, but at the moment when the correlation functions are called withing nfacctd.c these fields have not been yet decoded.

This PR adds a structs and functions to preload this information so that they are available when the correlation functions are called.

### Checklist

I have:

### DIFF
--- a/src/bgp/bgp_lookup.c
+++ b/src/bgp/bgp_lookup.c
@@ -28,7 +28,7 @@
 #include "pmbgpd.h"
 #include "rpki/rpki.h"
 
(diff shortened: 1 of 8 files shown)


paololucente (commented):
Thank you Leonardo ( @rodonile ) and sorry to have missed this went from draft to ready to be merged. I will review. Paolo

PR #712, opened 2023-08-02
pmtelemetryd: always activate TCP keepalive

ustorbeck:
TCP keepalive only gets activated on pure IPv4 systems. Activate it also on IPv4/v6 dual stack systems.

### DIFF
--- a/src/telemetry/telemetry.c
+++ b/src/telemetry/telemetry.c
@@ -357,8 +357,17 @@ int telemetry_daemon(void *t_data_void)
         Log(LOG_ERR, "ERROR ( %s/%s ): socket() failed. Terminating.\n", config.name, t_data->log_str);
         exit_gracefully(1);
       }
+    }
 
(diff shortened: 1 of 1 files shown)


ustorbeck:
Hi Paolo,

I assume TCP keepalive should always be activated on pmtelemetryd sockets, right?

At the moment keepalive only gets activated on pure IPv4 systems. On IPv4/v6 dual stack systems already [the first try to create the socket](https://github.com/pmacct/pmacct/blob/b7604af97286e1ffd3ac1e0aa55a21cf6b1e7961/src/telemetry/telemetry.c#L339) succeeds and [the block which sets the keepalive option](https://github.com/pmacct/pmacct/blob/b7604af97286e1ffd3ac1e0aa55a21cf6b1e7961/src/telemetry/telemetry.c#L343) is not executed.

This PR moves the instruction to set TCP keepalive out of the "if" block so that it's always activated and adds some error logging.

paololucente (commented):
Thanks Uwe ( @ustorbeck ), makes perfect sense! Merging. Paolo

PR #716, opened 2023-08-07
pmtelemetryd: free udp-notif buffer after reading data

ustorbeck:
fix a memory leak in pmtelemetryd when using udp-notif


ustorbeck:
rebased on current master

PR #719, opened 2023-08-24
Restore code to support vrfid:X as fallback RD

rodonile:
### Short description

Within the work in PR #698 I accidentally left out the fallback mechanism which sets the RD to "vrfid:X" where X=ingress_vrfid (or egress_vrfid respectively), for the situations where these fields are in the ipfix data but we are not getting the RD from options. This is useful especially to check if we are receiving correct information from the router.

This PR restores that functionality.

Cheers,
Leonardo

### Checklist

I have:

### DIFF
--- a/src/nfacctd.c
+++ b/src/nfacctd.c
@@ -3697,10 +3697,14 @@ void NF_init_zmq_host(void *zh, int *pipe_fd)
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thank you Leonardo ( @rodonile ), i will briefly review and merge. Paolo

PR #726, opened 2023-09-19
Implement flowset pre-processing checks (on end-of-flowset-padding)

rodonile:
### Short description
This PR introduces pre-processing checks on IPFIX Data records in nfacctd that can be enabled via a new config knob (nfacctd_pre_processing_checks). Currently the only check implemented is on the end-of-flowset padding, which is considered malformed if it is not composed of 0 bytes. In the future we might add other checks (e.g. on invalid timestamp formats or other fields).

The main purpose of this feature is to discard packets being decoded with a wrong template, which can happen if a router starts sending flow data encoded with a new template reusing older template IDs. This was sometimes observed on some buggy routers upon reboots/upgrades, where we received some data packets before receiving the updated templates. The result of this can be detrimental even with few packets, as all fields are wrongly decoded, delivering crazy numbers for e.g. packet and byte counts.

In general, according to RFCs 3954&7011, template ID reuse should be avoided. At least in any case it should be made sure that no data is being sent before sending out the respective template.

This feature gives us partial protection against some of these scenarios. 

### Checklist

I have:

### DIFF
--- a/CONFIG-KEYS
+++ b/CONFIG-KEYS
@@ -2105,6 +2105,19 @@ DESC:		This knob is to use the socket address instead of the one passed as part 
 		send Agent Address and that is used by default.
 DEFAULT:	false
 
+KEY:		[ nfacctd_pre_processing_checks ] [GLOBAL, ONLY_NFACCTD]
+VALUES:         [ true | false ]
+DESC:		This knob is to enable some pre-processing checks on the IPFIX/NFv9 Data Records in order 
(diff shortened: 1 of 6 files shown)


paololucente (commented):
Hi Leonardo ( @rodonile ), thanks for this useful contribution. At a first sight all is in order other than of course i see the usefulness of this. I will shortly review and merge. Paolo

PR #733, opened 2023-10-17
Add nvgre primitive for entity 351 layer2SegmentId with 0x02 MSB

rodonile:
### Short description
This PR defines a new primitive "nvgre", which represents the Tenant Network Identifier (TNI) for NVGRE L2 tunnels, which is encoded in IANA IE351 with MSB 0x02.

I opted to define a new variable (nvgre_tunnel_id) in the tunnel_primitives struct instead of reusing the existing tunnel_id which is currently used for vxlan. This way we can avoid having additional logic on the plugins side to determine which field to populate if both vxlan and nvgre are defined in the aggregate. Another possibility would have been extending the tunnel_id field to include also a tunnel_type identifier to distinguish between nvgre and vxlan, but this seemed a bit overkill.

Let me know if you agree with the current approach or you prefer another way. If we go this route we might even go ahead and rename tunnel_id with e.g. vxlan_tunnel_id for coherence. 

Cheers,
Leonardo

### Checklist

I have:


paololucente (commented):
Hi Leonardo ( @rodonile ), i agree with the approach you have preferred over the others. I will briefly review and merge. Paolo

PR #736, opened 2023-11-02
nfacctd: handle layer 2 flow information without IP addresses

ustorbeck:
This PR adds handling of layer 2 flow information without IP addresses to nfacctd which so far is ignored.
I've renamed the packet structure and removed the lookup functions according to your suggestions.

### DIFF
--- a/src/network.h
+++ b/src/network.h
@@ -626,6 +626,7 @@ struct cache_legacy_bgp_primitives {
 /* END: BGP section */
 
 struct packet_ptrs_vector {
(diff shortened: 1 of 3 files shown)


paololucente (commented):
Hi Uwe ( @ustorbeck ), thanks as always for this contribution! Very relevant and, while different in purpose, kind of in line with a feature that exists for sFlow (sfacctd), aggregate_unknown_etype. I will review & merge, leaving as future work to marry these feature together. Paolo

PR #740, opened 2023-11-06
rename bgp_attr->flag to rpki_maxlen

mxyns:
### Short description
`struct bgp_attr->flag` is only used in rpki for storing a prefix maximum length so let's rename the field to make this explicit :slightly_smiling_face:


paololucente (commented):
Thanks Maxence ( @mxyns ), i ruminated a bit on this one, although straightforward, yet to be fully convinced to have something named `rpki` in BGP attributes. Merging. Paolo

PR #742, opened 2023-11-21
Adding a config knob to exclude BMP stats from regular dump

rodonile:
### Short description
This PR adds the bmp_dump_exclude_stats config knob (false by default), which if enabled causes the BMP stats messages to not being cached, and thus not being included in the regular dump. This does not change the normal behavior, i.e. BMP stats are still streamed out by the plugins in real-time when received. 

The main reason why someone would enable this feature is to limit memory usage for situations where routers send a lot of BMP stats messages.

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ) for this useful contribution; i will briefly review and merge. Paolo

PR #752, opened 2024-01-16
Add bgp_id to output of BMP stats messages

arlake228:
### Short description

The changes in this pull request add the bgp_id field to the JSON and Avro output of BMP messages of type "stats". This field is included in the peer_up messages and policy messages already, this just includes it in the stats messages as well. The use case for adding it to the stats message is that in cases where we have multiple peerings, such as an IPv4 and an IPv6 peering, its useful to use the normalized bgp_id to filter/group the messages by router. We have this running against BMP data coming from our routers and seems to be working as advertised. This PR also updates the docs to reflect the change. 

### Checklist

I have:


paololucente (commented):
Hi Andrew ( @arlake228 ), thanks very much for this contribution. All looks in order at a glance; let me briefly review and test & i will merge. Paolo

PR #753, opened 2024-01-23
network.h: adds ifndefs to MAX_BGP_*_COMMS macros

sgg:
### Short description

There are cases where we may see more BGP communities that can fit in a 96 byte string. This diff makes the `MAX_BGP_STD_COMMS`, `MAX_BGP_EXT_COMMS`, and `MAX_BGP_LRG_COMMS` macros compile time configurable.

One could make this configurable at runtime, however that would be a far more invasive change and I haven't seen anyone asking for it.

### Checklist

I have:

### DIFF
--- a/src/network.h
+++ b/src/network.h
@@ -505,9 +505,15 @@ struct pkt_stitching {
 };
 
 /* START: BGP section */
+#ifndef MAX_BGP_STD_COMMS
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thanks Samani ( @sgg ), it makes perfect sense. Thanks for this contribution. Merging. Paolo

PR #757, opened 2024-02-14
Addressing pretag label overflow issue

rodonile:
This PR slightly adjusts the login in the pretag_copy_label() function to check if the null character is already in the dst string before adding one. This prevents the length to increase with multiple function calls, but also ensures that the pt_label_t struct will contain the correct length (label + null char).

### Checklist

I have:


paololucente (commented):
Great Leonardo ( @rodonile ), thanks very much for this much needed corrective patch. I have tested it at my end and found it working. Please release the draft flag so that i can merge it. Paolo

PR #758, opened 2024-02-15
Fix udp-notif message handling in pmtelemetryd

ustorbeck:
pmtelemetryd does not use the udp-notif library function to get the length of the payload buffer. It instead handles it as a string. This may lead to a buffer with additional characters added at the end which is not parsable and the content is dropped silently.

This PR fixes the udp-notif message handling and adds a few additional log messages.

### DIFF
--- a/src/telemetry/telemetry.c
+++ b/src/telemetry/telemetry.c
@@ -816,6 +816,8 @@ int telemetry_daemon(void *t_data_void)
               fd = TELEMETRY_UDP_NOTIF_FD;
             }
             else {
+              Log(LOG_WARNING, "WARN ( %s/%s ): consumer buffer size too small (%lu vs. %u bytes), dropping message\n",
+                  config.name, t_data->log_str, sizeof(consumer_buf), payload_len+1);
               unyte_udp_free_all(seg);
               goto select_again;
             }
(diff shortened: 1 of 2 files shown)


ustorbeck:
Hi Paolo
I'm not fully decided about the severity of the log message for the JSON errors. I would like to have it available in production environment, which usually does not have debug enabled (hence severity INFO). But then a device which sends faulty JSON messages may lead to flooding the logs, which would not happen with severity DEBUG. Do you prefer to reduce the severity of that message to DEBUG or shall I leave it at INFO level?

paololucente (commented):
Hi Uwe ( @ustorbeck ),

Thanks very much for this fix. With regard logs, you may use the log notification infrastructure defined in `log.c`, ie. log_notification_{set,isset,unset}. It is very easy in principle:
* you log something and attach it to a notification structure with a timeout;
* every time you have to log something related, check if the timeout is expired;
  - if positive, log again
  - if negative, hold up for this round

It will allow, say, to log once per minute or so instead that at every failing packet.

Paolo

ustorbeck:
Hi Paolo
Thanks for the hint. That throttling probably should be per peer then. As I do not understand the peer handling fully yet and to get the fix for the udp-notif buffer handling merged as soon as possible I've reduced the severity of the log message to DEBUG for now. I also can drop that commit from the PR if you prefer that.

paololucente:
Yes, that structure should be per peer. And ack about the change of severity to DEBUG. I'll merge in a second, we can revisit the logging part in a separate PR -- or i can take care of it for you. Paolo

paololucente:
Ack Uwe, let's keep it like that for now, sure. Paolo

PR #762, opened 2024-03-05
telemetry: fix timeout handling for udp-notif and grpc collectors

ustorbeck:
For udp-notif and grpc collectors the peer information is only partially deleted on timeout. When a peer starts to send data again after a timeout the data might either be assigned to the wrong peer or not forwarded to the output plugin at all. This PR fixes the timeout handling.


paololucente (commented):
Thanks for this Uwe ( @ustorbeck ), much appreciated. Paolo

Issue #766, opened 2024-03-12
[RPKI] RPKI version mismatch when TLS for RTR


SanderDelden:
**Description**
We are implementing insight into RPKI using PMACCT, when connecting our [Routinator](https://github.com/NLnetLabs/routinator) instance (which makes use of TLS for secure transport) we are getting an error referring to a version mismatch.

Example `nfacctd` configuration:
```
plugins: print[roa]

bgp_daemon: true
nfacctd_port: 5008
rpki_rtr_cache: 172.20.0.20:3323
rpki_rtr_cache_version: 0

aggregate[roa]: dst_roa
print_output_file[roa]: /tmp/pmacct/1m_roa.json
print_output[roa]: json
print_history[roa]: 1m
print_history_roundoff[roa]: m
print_refresh_time[roa]: 60
print_output_file_append[roa]: true
```

Example `docker-compose.yml`:
```
%YAML 1.1
---
version: '3.8'
services:
  nfacctd:
    image: pmacct/nfacctd:bleeding-edge
    volumes:
    - $PWD/pmacct.conf:/etc/pmacct/nfacctd.conf:ro
    networks:
      pmacct_network:
        ipv4_address: 172.20.0.10

  routinator:
    image: nlnetlabs/routinator:v0.13.2
    command: [
    "server",
    "--rtr-tls", "0.0.0.0:3323",
    "--rtr-tls-cert", "/home/routinator/ssl/rtr/routinator-rtr.crt",
    "--rtr-tls-key", "/home/routinator/ssl/rtr/routinator-rtr.key",
    "--http-tls", "0.0.0.0:8323",
    "--http-tls-cert", "/home/routinator/ssl/http/routinator-http.crt",
    "--http-tls-key", "/home/routinator/ssl/http/routinator-http.key"
    ]
    volumes:
    - $PWD/ssl:/home/routinator/ssl:ro
    networks:
      pmacct_network:
        ipv4_address: 172.20.0.20

networks:
  pmacct_network:
    ipam:
      config:
      - subnet: 172.20.0.0/24
```

When starting the containers the following log entries are observed:
```
INFO ( default/core/RPKI ): Connected to RTR Cache: 172.20.0.20:3323
WARN ( default/core/RPKI ): rpki_rtr_parse_msg(): RPKI version mismatch (me=0 cache=21)
```

If we disable the TLS for RTR the error is no longer observed.

**Version**
[...]

PR #767, opened 2024-03-12
Add remark regarding RTR-over-TLS to CONFIG-KEYS

SanderDelden:
Addresses #766


job:
Perhaps change it to “only unprotected transport over TCP is supported”. Listing everything that’s not supported would make the document long

paololucente (commented):
Thanks Sander ( @SanderDelden ) for this clarification, much appreciated. Paolo

PR #771, opened 2024-03-19
ci: use shallow git clones and avoid intermediate files when downloading build dependencies

ustorbeck:
this uses less space and gains a bit of speed when building pmacct


paololucente (commented):
Thank you Uwe ( @ustorbeck ) for this. Paolo

PR #773, opened 2024-03-21
Fixes and Improvements in the BMP/BGP High Availability Feature

rodonile:
### Short description
This PR solves some issues add adds some additional configuration possibilities on the HA feature and the redis library. 

**Changelog**:
- new config flags to set HA-specific cluster parameters (to differentiate from cluster_id, cluster_name that were already being used by other functionalities such as eBPF)
- new config flags to set a custom value for the queue message timeout and queue max size
- bind redis to v1.2.0 in deps.sh script for building docker images
- extend README for HA and provide deployment example
- add logging of the daemon state (i.e. ACTIVE or STAND-BY) every 10mins
- add redis connection and redis command timeouts to avoid long stalls when connection with redis server is going down abruptly (without TCP reset)

### Checklist

I have:


paololucente (commented):
Thanks Leonardo ( @rodonile ) for taking ownership of this much useful future. Merging. Paolo

PR #793, opened 2024-07-09
Add "nfacctd_account_options: true" to nfacctd.conf example 

Bell-Charles:
This pull request adds the following:

1.  Update CONFIG-KEYS to add line "nfacctd_account_options: true" to the nfacctd.conf example for nfacctd_account_options in CONFIG-KEYS
2. Add support of ingress vrf id to vrf name mapping so that the vrf name can be shown instead of the vrf id in the json


paololucente (commented):
Hi Charles ( @Bell-Charles ), thanks very much for these two contributions -- much appreciated. As we discussed, could you please split them. Either close this PR and open two new (probably the easiest) or do some other magic to make it happen? Thanks, Paolo

PR #797, opened 2024-07-16
telemetry: fix message loss in GRPC collector mode

ustorbeck:
pmtelemetryd loses messages when running in GRPC collector mode (with --enable-grpc-collector)


paololucente (commented):
Thanks as always Uwe ( @ustorbeck )

PR #814, opened 2024-09-06
Fix excessive BGP table deletes triggered by BMP loc-rib peer_down messages

rodonile:
### Short description

This PR addresses the following:

- adding the **peer_distinguisher** field in the bgp_peer structure to be used when describing a BGP peer learned via BMP. This is necessary to be able to distinguish loc-rib peers, which all have address 0.0.0.0 and need to be distinguished via the peer_distinguisher field in the BMP per-peer header (which matches the RD of the VRF in the router originating the BMP session). I added new compare functions used to add/search/delete BGP peer structures in the tree for BMP.

- adjusting the peer lookup function for BMP (**bgp_lookup_find_bmp_peer**) to only fallback to remote peers when we are sure that no BMP peer matching the socket addr for the flow record is found. Otherwise we might be using incomplete information for correlation

- adding a test case (403) including such loc-rib peer-down messages to validate the changes

### Checklist

I have:


paololucente (commented):
Thank you Leonardo ( @rodonile ) for this very useful contribution. Given the extent of the changes, I'll review and test a bit at my end too before merging. Paolo

marcotollini:
Ciao @paololucente, did you have a chance to test this? :)

paololucente:
Ciao Marco ( @marcotollini ), still ongoing but i am on it. Please allow me a few more days.

PR #816, opened 2024-09-13
actions: decouple regression tests from ci.yaml

msune:
More prep work for ARM support:

> Modularize regression tests in a dedicated reusable workflow. Use a shared composite action to build pmacct containers. This makes both workflows shorter and easier to debug.
>
> In regression_tests workflow, step 'hub-pulled-docker' is renamed to 'cache-docker-images' and pmacct is compiled in there using the composite action.

### Checklist

I have:


paololucente (commented):
Thanks Marc ( @msune ) for preparing the field for ARM support. And of course improved modularization is always a pro. Leonardo ( @rodonile ), would you mind giving this a look and blessing from your side too? To me it all looks good. Paolo

rodonile:
Hi @paololucente, @msune. All good from my side, makes the whole workflows cleaner. Thanks, Leonardo.

PR #821, opened 2024-09-30
ci: fix/revert uploading of images job

msune:
Commit 91df0c752, part of #774, broke `publish-dockerhub` job by changing the run condition, which makes it skip runs that should otherwise run.

```
  runs-on: ubuntu-22.04
- if: github.event_name != 'pull_request'
+ if: ${{ github.event_name == 'schedule' && github.repository == 'pmacct/pmacct' }}
```

Reverting this part of 91df0c752.

---

@paololucente @rodonile hopefully this won't happen anymore thanks to #816. But please do NOT conflate these changes in unrelated commits, else these things slip in unnoticed - especially if commits are as big as 91df0c752 is:

https://github.com/pmacct/pmacct/commit/91df0c7529266467a99a0d02c042f19bbffda22e#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL162

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -147,7 +147,7 @@ jobs:
   publish-dockerhub:
     needs: [pmacct-docker, build-and-test]
     runs-on: ubuntu-22.04
-    if: ${{ github.event_name == 'schedule' && github.repository == 'pmacct/pmacct' }}
+    if: github.event_name != 'pull_request'
     env:
       DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
       DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}


msune:
> Hi @msune. I changed that to allow the workflows to run on forks without failing for that single job.

@rodonile this diff https://github.com/pmacct/pmacct/commit/91df0c7529266467a99a0d02c042f19bbffda22e#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL162

doesn't preclude a workflow to run on a fork. In fact, I am testing it in my fork all the time with my own dockerhub account mimic pmacct ones to test.

> The job is not broken, it's still running every night with the nightly cron, just not running at every single commit.

And **[by definition](https://github.com/pmacct/pmacct/blob/master/docs/DOCKER.md#tags)** this means it's broken; bleeding edge HAS to upload on the HEAD of **every commit** merged in pmacct as per documentation:

>> _bleeding-edge: only for the brave. Latest commit on master. This container is not recommended to be used in production._

Moreover, if we tag something midday (say `1.7.11`), and we just push anything before midnight, the release container will not only not be generated on time (read, the moment we push the tag), it will just NOT be generated, because you will never run a CI workflow with that commit in the HEAD. **This breaks docker container publishing completely...** So please, be careful, and split changes as important as this ones not only in separate commits, but in PRs in itself.

> I should have mentioned it in the PR notes, sorry for that. Do you have something against keeping the github.repository check though? (we can configure it to run at every commit as well if you were relying on that)

Yes, this is wrong, because I can't test it on my fork without patching it.

What is the problem you are trying to fix? If the problem is that this fails in your fork, and you don't want to run it, we can introduce a conditional env. variable (e.g. `SKIP_DOCKERHUB_PUBLISH` or alike), which you or any developer can set to `1` in **your fork**. 

This would _a_  correct way to solve your problem (not the only one), but certainly not https://github.com/pmacct/pmacct/commit/91df0c7529266467a99a0d02c042f19bbffda22e and certainly not silently including it as part of a massive PR without explicit discussion.

rodonile:
> Yes, this is wrong, because I can't test it on my fork without patching it.

Makes sense, I see. 

> What is the problem you are trying to fix? If the problem is that this fails in your fork, and you don't want to run it, we can introduce a conditional env. variable (e.g. SKIP_DOCKERHUB_PUBLISH or alike), which you or any developer can set to 1 in your fork.

Yep, this is what I was addressing. Let's do it with the env variable, I like the idea.

msune:
>Yep, this is what I was addressing. Let's do it with the env variable, I like the idea.

Please, issue a PR after this one is merged. I will review it. We need to fix the publishing of images ASAP.

@paololucente please merge this PR whenever you can, and let's see that bleeding-edge is uploaded as it should.

msune:
In all fairness, and notwithstanding any of what was said above, I myself also introduced a regression in publishing the docker images in #816, see #822.

We all need to be more careful...

PR #823, opened 2024-10-02
CI: add env var to allow switching off dockerhub publish job

rodonile:
### Short description
As discussed with @msune , this PR adds a config env variable that can be declared in the actions variables to skip the dockerhub image publishing job. 

### Checklist

I have:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -162,7 +162,7 @@ jobs:
   publish-dockerhub:
     needs: [pmacct-docker, build-and-test]
     runs-on: ubuntu-22.04
-    if: github.event_name != 'pull_request'
+    if: github.event_name != 'pull_request' && vars.SKIP_DOCKERHUB_PUBLISH != 'true'
     env:
       DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
       DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}


.github/workflows/ci.yaml
```diff
@@ -162,7 +162,7 @@ jobs:
   publish-dockerhub:
     needs: [pmacct-docker, build-and-test]
     runs-on: ubuntu-22.04
-    if: github.event_name != 'pull_request'
+    if: github.event_name != 'pull_request' && vars.SKIP_DOCKERHUB_PUBLISH != 'true'
```
msune, line 165: Is `vars` is the right context for repository-wide variables? I am not familiar with `vars` (introduced 2023), but in the past I used `env`, which has the advantage that you can dynamically adjust its value at runtime as a result of previous steps/jobs execution.

rodonile: I also just found out about this, but vars is the context that you can use to access repository variables defined in the repo settings (they also call them configuration variables in the docs). This way we can very easily add the variable there (https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#creating-configuration-variables-for-a-repository).

Also according to documentation then we could bring it in via:

```
env:
  # Setting an environment variable with the value of a configuration variable
  env_var: ${{ vars.ENV_CONTEXT_VAR }}
```

if we know we'd need to dynamically adjust the value. Are you thinking about other situations where this job needs to be skipped?

msune: I am talking about dynamically changing the value from one step or even job to another using GITHUB_ENV + GITHUB_OUTPUTS

But it doesn't matter for now, if we need that we can refactor it.

ACK

PR #826, opened 2024-10-08
CI/Regression Tests: add pmtelemetryd support and yang udp-notif and tcp-json test cases

rodonile:
### Short description
@msune @paololucente  this is still work in progress but I'm opening a draft PR since I'd like to discuss some points.

This PR adds pmtelemetryd support within the regression testing framework and adds 2 test with Yang telemetry data. One with udp-notif and one with cisco proprietary tcp-json transport.

For udp-notif the _--enable-unyte-udp-notif_ config flag is not present in the base image. Currently I have just added it to the base Dockerfile, but I'd like to discuss with you how you prefer this to be handled. Consider that in the future we will also need to create an image for pmtelemetryd+mdt-dialout-collector for grpc transport.

Two options:
1. add the _--enable-unyte-udp-notif_ config flag to the base Dockerfile, which allows us to use the pmtelemetryd image for both udp-notif and tcp-json cases. 
2. add another Dockerfile for **pmtelemetryd_udp_notif** which does not use the base Dockerfile and has the additional config flag, then add an action to build it, which can be called by the regression_tests and also by the ci workflows if we also want to push it to Dockerhub. For grpc we will also most likely do the same in the future, i.e. building **pmtelemetryd_grpc** image.

What do you prefer? Do you have other ideas?

Cheers,
Leonardo

### Checklist

I have:


msune:
> For udp-notif the --enable-unyte-udp-notif config flag is not present in the base image. Currently I have just added it to the base Dockerfile, but I'd like to discuss with you how you prefer this to be handled

IMO, the preference should be adding it to the base container

PR #835, opened 2024-10-29
fix, BMP daemon: mem leak related to bmp_rib_type_print() usage - part 2

rodonile:
### Short description
Commit https://github.com/pmacct/pmacct/commit/a681b10802753183c602457a7be8781119103bfc introduced some memory leaks which were partially fixed by https://github.com/pmacct/pmacct/commit/4606091087a20c977bad76b27e096e93bde6fd76, but 2 calls of the bmp_rib_type_print() function were not considered in the fix.

### Checklist

I have:


paololucente (commented):
Thanks Leonardo ( @rodonile ) for this supplement to the previous, incomplete, patch. Much appreciated as always. Paolo

PR #837, opened 2024-11-03
Write Peer Distinguisher into separate field

### LINKED ISSUE #783: How to discern BMP instance peer distinguisher and route distinguisher if both are present?
**Description**
As I discovered `bmp` daemon outputs `Peer Distinguisher` (PD) from `Per-Peer` header as `rd` and `rd_origin` fields of dumping object. However, `bgp` route also may have a route distinguisher (RD) with the same output field names.

So I am afraid, then in case if both PD and RD are present there would be a little mess inside `bmp` daemon output.

Also I would suggest to rename `Peer Distinguisher` output name to `pd`. 

We use 1.7.6 version of `pmacct`

WBR

GeorgyKirichenko:
BMP protocol header contains peer distinguisher field which is reported by `bmpd` as `rd`/`rd_origin` inside json/avro/etc output. However a route may contain route distinguisher reported in the same output.

The patch introduces `pd`/`pd_origin` output for `bmpd` in order to discern these two values.

Closes #783

### DIFF
--- a/src/bmp/bmp_logdump.c
+++ b/src/bmp/bmp_logdump.c
@@ -413,11 +413,11 @@ int bmp_log_msg_stats(struct bgp_peer *peer, struct bmp_data *bdata, struct pm_l
     }
 
     if (!is_empty_256b(&bdata->chars.rd, sizeof(bdata->chars.rd))) {
-      char rd_str[SHORTSHORTBUFLEN];
+      char pd_str[SHORTSHORTBUFLEN];
 
-      bgp_rd2str(rd_str, &bdata->chars.rd);
-      json_object_set_new_nocheck(obj, "rd", json_string(rd_str));
-      json_object_set_new_nocheck(obj, "rd_origin", json_string(bgp_rd_origin_print(bdata->chars.rd.type)));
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Hi @GeorgyKirichenko ,

Thanks very much for this contribution and addressing the issue you have opened earlier on.

Question: people may have pmbmpd in production right now and expect the current behaviour -- for good or for bad. Could you make this optional, ie. make it part of a config option? You could use as inspiration this commit: https://github.com/pmacct/pmacct/commit/7344220b75c2f3da85898e5f82e990728fb5542f

If you think it will take too much of your time, no worries, i can merge the PR as it is and work out the config knob myself. Just let me know.

Paolo

GeorgyKirichenko:
Hi @paololucente ,
glad to hear this from you!

Good catch about changing existing behavior, I will fix this soon and update the PR

Have a nice day!

paololucente:
Closing this without merging as PR #888 supersedes this one.

paololucente (commented):
Closing this without merging as PR https://github.com/pmacct/pmacct/pull/888 supersedes this one.

PR #844, opened 2024-12-05
Implementation of vrf_name primitive

Bell-Charles:
### Short description
1.  the vrf_name aggregate keyword, this keyword sets vrf_name to the followig:
    - vrf_name is set to vrf_name mapped to ingress_vrf_id
    - if ingress_vrf_id is 0 or is “default” and egress_vrf_id is mapped to a vrf_name that is not “default”, then vrf_name is set to the vrf name mapped to egress_vrf_id

2.  fixed the avero schema so a empty string is always added to the data output

3.  Also made sure the array holding vrf names are allocated to the right size

### Checklist

I have:
- [ x] compiled & tested this code
- [ x] included documentation (including possible behaviour changes)


paololucente (commented):
Hi Charles ( @Bell-Charles ),

Thanks for this contribution. I will review and merge. Two questions:
* What is the role of USE_VRF_NAME_PTR? Is it needed?
* is `vrf_name` implemented on top of ingress/egress vrf name primitives? I thought it would be in place of these.

We can discuss further offline.

Paolo

Bell-Charles:
The USE_VRF_NAME_PTR was suggested in my last pull request in case we want to avoid creating a static array to store the vrf name in packet_ptrs in the future.   I could delete this if you think this is not needed.

Actually vrf_name and ingress_vrf_name/egress_vrf_name are both needed for different use cases.   

P.S.  It looks like some of the test cases uses vrf_name as a custom_primitive that now fails.  Do you think this is because my new primitive conflict with this custom primitive?

rodonile:
It would be better to fix the tests before merging. Yes, reason is that  vrf_name was present as custom primitive for some of the tests

Bell-Charles:
I agree we should fix the tests.   Any idea how I can address the name collision between custome_primitive and the new primitive.  Should I just change the name in the test cases?

Bell-Charles:
I have put the vrf_name back in the unit-test

PR #851, opened 2025-01-17
Fix inconsistency of command line options

corubba:
### Short description

While playing around with the code, I noticed an inconsistency in the cli options. So I threw some quick-and-hacky scripting at the sourcecode to compare the `ARGS_*` defines in `pmacct-defines.h` against the `usage_*` functions and `optstr()`-switch-cases in the executables main C files, and from there proceeded to cause-research and fixing:

* `uacctd`'s `-N` was removed in 445f96cd but still present in the optstring.
* `pmacct`'s `-t` was defined in the optstring with an required argument, but it takes none. The argument was wrongly "inherited" in 45c860a776 by the removal of `-m`.
* `pmacct`'s `-P` was missing in the usage output since the beginning in 1f7b559c. Because it's unused since f78fe098, remove the option completely.
* `pmtelemetryd`'s `-Z` was removed in e3cce045 but still present in the usage output.

### Checklist

I have:


paololucente (commented):
Thanks very much @corubba for your eye to detail, much appreciated. Paolo

PR #853, opened 2025-01-23
tee plugin: reporting tee_send execution count per receiver

strzebon:
This Pull Request introduces monitoring of how many times a message has been sent to a specific receiver by tee plugin. The counters are periodically exported to .csv file. Example structure of the file:
```
POOL_ID,RECEIVER_IP,RECEIVER_PORT,TEE_SENT,REFRESH_TIME
0,127.0.0.1,1234,10000,60
1,127.0.0.2,1234,10000,60
```
Two new config properties were added:
`tee_counters_refresh_time` - time interval in seconds at which the counters are reported to file and reset (optional, default 60s)
`tee_counters_path` - file where counters are reported to (if not provided, counters are not reported)

### DIFF
--- a/src/cfg.c
+++ b/src/cfg.c
@@ -449,6 +449,8 @@ static const struct _dictionary_line dictionary[] = {
   {"tee_ipprec", cfg_key_nfprobe_ip_precedence},
   {"tee_pipe_size", cfg_key_tee_pipe_size},
   {"tee_kafka_config_file", cfg_key_tee_kafka_config_file},
+  {"tee_counters_refresh_time", cfg_key_tee_counters_refresh_time},
+  {"tee_counters_path", cfg_key_tee_counters_path},
   {"bgp_daemon", cfg_key_bgp_daemon},
   {"bgp_daemon_ip", cfg_key_bgp_daemon_ip},
   {"bgp_daemon_interface", cfg_key_bgp_daemon_interface},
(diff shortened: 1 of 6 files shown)


paololucente (commented):
Hi Witold ( @strzebon ), thanks very much for this contribution.

May i ask you a couple of related questions:

1) Can you mention a use-case for this feature? I see its usefulness but was curious how you are using it, whether manually checking the data or a 3rd party tool checks the file; 

2) Have you thought of sending this output out to the log, instead that to a specific file? Especially going through syslog, a number of options and features become available for free (including file rotation, for example). A reader could still easily filter / grep through the content; maybe in that case a JSON output (or anything self-describing) would be more favorable than CSV that needs the title to be parsed first.

Paolo

wydrych:
Our idea was to keep it as similar to the print plugin as possible to keep the things generic for whoever parses the CVS anyhow :-)

1. The use case is to have a generic monitoring framework for the tee plugin. We're forwarding samples to various destinations, sometimes with filtering. While we have metrics at each destination, monitoring the tee plugin helps identify issues at each stage.
2. We saw it being similar to the print plugin – CSV may not be ideal, but it is generic and easy to parse or process further into other monitoring systems.

wydrych:
Hi Paolo @paololucente,

ping :-) Do you think we could have it merged?

strzebon:
Hi Paolo @paololucente,

I've updated this PR. When you have a chance, could you take another look? If everything looks good, would it be possible to merge it?

Thanks!

PR #856, opened 2025-02-09
Improve handling of unknown map keys

corubba:
# Short description
Improve the way unknown map keys are reported/logged during map loading, making it easier and more obvious if/when things are wrong. This is done by improving the log messages and the info they provide, as well as reporting it as an error if the key is known but not supported by the current daemon.

# Long description

To demonstrate the issue I am trying to solve, assume the following two config files, `uacctd.conf` and `pre-tag.map`. These make uacctd do a simple aggregation by source, add a label, and print the result to stdout.
```
uacctd_group: 30

pre_tag_map: pre-tag.map

aggregate: src_host, label

plugins: print

print_output: csv 
print_output_file: /dev/stdout
print_refresh_time: 3
```
```
set_label=rule src_net=255.255.255.255/32
set_label=catchall
```

Running `uacctd -f uacctd.conf` will produce the following console output:

```
INFO ( default/core ): Linux NetFilter NFLOG Accounting Daemon, uacctd (20250121-2 (912fb121))
INFO ( default/core ):  '--enable-nflog' '--disable-shared' '--enable-debug' '--enable-l2' '--enable-traffic-bins' '--enable-bgp-bins' '--enable-bmp-bins' '--enable-st-bins'
INFO ( default/core ): Reading configuration file '/tmp/pmacct/uacctd.conf'.
INFO ( default_print/print ): plugin_pipe_size=4096000 bytes plugin_buffer_size=1284 bytes
INFO ( default_print/print ): ctrl channel: obtained=212992 bytes target=25520 bytes
INFO ( default/core ): [pre-tag.map] (re)loading map.
INFO ( default_print/print ): cache entries=16411 base cache memory=56322552 bytes
INFO ( default/core ): [pre-tag.map] map successfully (re)loaded.
INFO ( default/core ): [pre-tag.map] (re)loading map.
INFO ( default/core ): [pre-tag.map] map successfully (re)loaded.
INFO ( default/core ): Successfully connected Netlink NFLOG socket
INFO ( default_print/print ): *** Purging cache - START (PID: 69470) ***
LABEL,SRC_IP,PACKETS,BYTES
rule,127.0.0.1,2,168
INFO ( default_print/print ): *** Purging cache - END (PID: 69470, QN: 1/1, ET: 0) ***
INFO ( default/core ): SIGINT received. Exiting ...
```

Please note that [`src_net` is not supported in uacctd pre-tag-maps](https://github.com/pmacct/pmacct/blob/912fb121c5099d83b637e3f29bfc316f065b2a97/examples/pretag.map.example#L25), but there are no warnings or errors regarding the pre-tag-map containing unknown or unsupported keys.
Also note that the flow is labeled with `rule` and not `catchall` despite it not matching the given `src_net` key. This is because for acct-type `ACCT_PM` (which is used by uacctd), the [`PT_map_src_net_handler()`](https://github.com/pmacct/pmacct/blob/912fb121c5099d83b637e3f29bfc316f065b2a97/src/pretag_handlers.c#L1394) adds no function to the map entry but also returns `FALSE` (aka "no error"), effectively skipping the `src_net` match and the loaded rule/entry only consists of the `set_label=rule` key, which then (of course) is matched by every flow.
I am unsure why [`load_plugins()`](https://github.com/pmacct/pmacct/blob/912fb121c5099d83b637e3f29bfc316f065b2a97/src/plugin_hooks.c#L332) loads the pre-tag-map twice while there is only one plugin, but that's research for another day.

After applying the changes from this PR, the following console output will be produced:

```
INFO ( default/core ): Linux NetFilter NFLOG Accounting Daemon, uacctd (20250209-2 (efd7860))
INFO ( default/core ):  '--enable-nflog' '--disable-shared' '--enable-debug' '--enable-l2' '--enable-traffic-bins' '--enable-bgp-bins' '--enable-bmp-bins' '--enable-st-bins'
INFO ( default/core ): Reading configuration file '/tmp/pmacct/uacctd.conf'.
INFO ( default_print/print ): plugin_pipe_size=4096000 bytes plugin_buffer_size=1284 bytes
INFO ( default_print/print ): ctrl channel: obtained=212992 bytes target=25520 bytes
INFO ( default/core ): [pre-tag.map] (re)loading map.
INFO ( default_print/print ): cache entries=16411 base cache memory=56322552 bytes
WARN ( default/core ): [pre-tag.map:1] Line ignored because of unknown key 'src_net'.
INFO ( default/core ): [pre-tag.map] map successfully (re)loaded with 1 entries.
INFO ( default/core ): [pre-tag.map] (re)loading map.
WARN ( default/core ): [pre-tag.map:1] Line ignored because of unknown key 'src_net'.
INFO ( default/core ): [pre-tag.map] map successfully (re)loaded with 1 entries.
INFO ( default/core ): Successfully connected Netlink NFLOG socket
INFO ( default_print/print ): *** Purging cache - START (PID: 70902) ***
LABEL,SRC_IP,PACKETS,BYTES
catchall,127.0.0.1,2,168
INFO ( default_print/print ): *** Purging cache - END (PID: 70902, QN: 1/1, ET: 0) ***
INFO ( default/core ): SIGINT received. Exiting ...
``` 

There are now warnings that the `src_net` key in line 1 of the pre-tag-map is unknown (aka unsupported), and that the respective line/rule is ignored. You can now also see that the loaded map contains only one entry/rule, and that the flow is correctly labeled `catchall`.

# Commits

## Return NotFound for unimplemented pre-tag-map keys
When there is no implementation for the current key, instead of making it a no-op and return ok, return an NotFound error which a) gets logged and b) makes the whole line/rule getting ignored.

## Make unknown key log message more specific
The original log messages makes it sound like only the unknown key is ignored, but actually the whole line/rule is.

## Log map size after (re-)load
After successfully (re-)loading a map, also log the number of entries it then has. This makes it possible to sanity-check if the map was loaded as expected.

# Checklist

I have:

### DIFF
--- a/docs/SIGNALS
+++ b/docs/SIGNALS
@@ -47,7 +47,7 @@ SIGUSR2:	if 'maps_refresh' config directive is enabled, it causes maps
 		be derived by the logs, ie.:
 
 		INFO ( default/core ): [/path/to/primitives.lst] (re)loading map.
-		INFO ( default/core ): [/path/to/primitives.lst] map successfully (re)loaded.
+		INFO ( default/core ): [/path/to/primitives.lst] map successfully (re)loaded with 3 entries.
 
 		primitives.lst is loaded by the core process and a signal to
 		reload such map should be sent to the Core Process. There is
--- a/src/pretag.c
+++ b/src/pretag.c
@@ -230,7 +230,7 @@ void load_id_file(int acct_type, char *filename, struct id_table *t, struct plug
                     else err = E_NOTFOUND; /* key not found */
                   }
                   if (err) {
-                    if (err == E_NOTFOUND) Log(LOG_WARNING, "WARN ( %s/%s ): [%s:%u] unknown key '%s'. Ignored.\n",
+                    if (err == E_NOTFOUND) Log(LOG_WARNING, "WARN ( %s/%s ): [%s:%u] Line ignored because of unknown key '%s'.\n",
 						config.name, config.type, filename, tot_lines, key);
                     else Log(LOG_WARNING, "WARN ( %s/%s ): [%s:%u] Line ignored.\n", config.name, config.type, filename, tot_lines);
                     break;
@@ -246,7 +246,7 @@ void load_id_file(int acct_type, char *filename, struct id_table *t, struct plug
                     else err = E_NOTFOUND; /* key not found */
                   }
                   if (err) {
-                    if (err == E_NOTFOUND) Log(LOG_WARNING, "WARN ( %s/%s ): [%s:%u] unknown key '%s'. Ignored.\n", 
+                    if (err == E_NOTFOUND) Log(LOG_WARNING, "WARN ( %s/%s ): [%s:%u] Line ignored because of unknown key '%s'.\n",
 						config.name, config.type, filename, tot_lines, key);
                     else Log(LOG_WARNING, "WARN ( %s/%s ): [%s:%u] Line ignored.\n", config.name, config.type, filename, tot_lines);
                     break;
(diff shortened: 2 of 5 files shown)


paololucente (commented):
Hi @corubba ,

Thanks for this contribution, it indeed does make a lot of sense! I'll test it briefly at my end then merge.

Paolo

PR #857, opened 2025-02-11
Fix buffer corruption in vrf_name aggregate

Bell-Charles:
### Short description
Add PmLabelTSz to check_pipe_buffer_space(), otherwise the buffer corruptions are seen
Also included in the changes are:

1.  Don't add to the buffer if vrf_name is empty
2.  Only allocate enough buffer to hold actual length of the vrf_name instead of max size

### Checklist

I have:

### DIFF
--- a/src/pkt_handlers.c
+++ b/src/pkt_handlers.c
@@ -2043,11 +2043,15 @@ void NF_ingress_vrf_name_handler(struct channels_list_entry *chptr, struct packe
 {
   struct pkt_vlen_hdr_primitives *pvlen = (struct pkt_vlen_hdr_primitives *) ((*data) + chptr->extras.off_pkt_vlen_hdr_primitives);
 
-  if (check_pipe_buffer_space(chptr, pvlen, MAX_VRF_NAME)) {
+  if (strlen(pptrs->ingress_vrf_name) == 0) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thanks very much @Bell-Charles for the quick fix! Much appreciated & merging.

src/pkt_handlers.c
```diff
@@ -2043,34 +2043,47 @@ void NF_ingress_vrf_name_handler(struct channels_list_entry *chptr, struct packe
 {
   struct pkt_vlen_hdr_primitives *pvlen = (struct pkt_vlen_hdr_primitives *) ((*data) + chptr->extras.off_pkt_vlen_hdr_primitives);
 
-  if (check_pipe_buffer_space(chptr, pvlen, MAX_VRF_NAME)) {
+  if (strlen(pptrs->ingress_vrf_name) == 0) {
+    return;
+  }
+
+  if (check_pipe_buffer_space(chptr, pvlen, PmLabelTSz+strlen(pptrs->ingress_vrf_name)+1)) {
```
msune, line 2050: I didn't see it before; `strlen()` is expensive. 

Might have been a good idea to define:

```
int ingress_vrf_name_len = strlen(pptrs->ingress_vrf_name) +1;
```

Change `if()` clause to ` == 1` and use the stack var everywhere else.
And change the 

And immed

PR #866, opened 2025-03-28
Add libzstd to docker base image

floatingstatic:
### Short description
Adds `libzstd` to the docker base image to support the use of `zstd` kafka compression. 

### Checklist

I have:

### DIFF
--- a/ci/Dockerfile-ubuntu-focal-for-pmacct
+++ b/ci/Dockerfile-ubuntu-focal-for-pmacct
@@ -25,6 +25,7 @@ RUN apt-get update && \
     libpcap-dev \
     libpq-dev \
     libsnappy-dev \
+    libzstd-dev \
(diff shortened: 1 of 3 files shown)


floatingstatic:
For reference this is what I get when setting `global, compression.codec, zstd` in the librdkafka config with current upstream containers in docker hub:

```
2025-03-28T14:03:26Z WARN ( raw-dev-awsuse2/kafka ): [/etc/sfacctd/kafka.cfg:18] key=compression.codec value=zstd failed: Unsupported value "zstd" for configuration property "compression.codec": libzstd not available at build time
```

paololucente (commented):
Thanks Jeremiah ( @floatingstatic ) for this, if we support snappy in the base image, we can indeed support further compression methods. Merging. Paolo

PR #879, opened 2025-07-27
sfacctd: accept sflow packets with dropped packet notifications.

yzp0n:
Before this change, sfacctd would drop the sflow packet entirely if it contained a sample type of `enterprise = 0; format = 5`, which corresponds to a dropped packet notification [1].

This change allows sfacctd to accept such packets and ignore the dropped packet samples, thus allowing it to continue processing other valid samples in the same packet.

The change would also make the `sfacctd_disable_sanity_checks` option apply to unknown sflow v5 sampleTypes.

[1] https://sflow.org/sflow_drops.txt

### Checklist

I have:
    - N/A: No new files are added in this PR
- [ ] included documentation (including possible behavior changes)


yzp0n:
For additional context, the use case I have is to allow ingestion from VPP + host-sflow generated sflow to feed into my flow analyzer pipeline which coexists with the pmacct promiscuous captures on non-VPP routers, while processing dropped packets on other processor utils in-house.

yzp0n:
Thanks for the CI run. I believe I've fixed the error by removing the unused variable. Would you mind running the CI again?

paololucente (commented):
Thanks @yzp0n for this contribution. Makes sense to skip dropped samples and have the option to not discard sFlow packets in presence of an unknown sample type. Merging. Paolo

PR #885, opened 2025-08-07
test: add pytest+scapy test framework+first test

msune:
This commit:

* Adds a test/ folder. This folder contains a pytest infrastructure that can run tests using `make check`
* A simple `nfacctd` test which:
  - Starts nfacctd with the JSON file plugin. It uses the local built binary.
  - Wait for the first PURGE trace in `nfacctd` (to avoid races in the test, so that the pcap sniffer is "ready")
  - Injects a nfv9 template pkt (using loopback addresses)
  - Injects a nfv9 flowlog (single, but the test is tunable)
  - It reads the JSON file and compares with the injected data.

This test appears to fail 3/4 executions in my local machine (75% of the time), and it seems to be logically "correct".

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,6 +8,9 @@ if USING_SQL
 EXTRA_DIST += src/sql_common_m.c
 endif
 
+# Add test subdir at the very end
+SUBDIRS += test/
+
 pmacct_data_dir = $(datarootdir)/pmacct
 pmacct_libs_dir = $(libdir)/pmacct
 pmacct_docs_dir = $(pmacct_data_dir)/docs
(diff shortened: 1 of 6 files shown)


msune:
Draft PR for discussion. Still not integrated with the CI

msune:
@paololucente did you have the chance to look into this failure so that we can decide whether to merge this initial test?

paololucente (commented):
hey Marc ( @msune ), for me this is good to merge and a good addition to the test framework. I don't spot any failures really, please go ahead and remove the draft tag to merge. Kudos and thanks as always! Paolo

PR #892, opened 2025-10-14
Introducing a new cap_4as struct and exposing 2 static functions

Pedro-W21:
### Short description
This pull request :
- changes the cap_4as field on bgp_peer structs from a pointer to a struct containing a boolean and a u_int32_t representing the as4
- exposes 2 previously static functions in bgp_ecommunity.c and bgp_lcommunity.c

### Checklist
I have:


paololucente (commented):
Hi Pierre ( @Pedro-W21 ), Thanks for this contribution, it makes a lot of sense. May i ask you for two minor style changes before merging?

* Use `TRUE` instead of `true`;
* Change name of `used` into `is_used`

Paolo

PR #894, opened 2025-10-15
fix for new bmp_daemon_set_pd config and "pd" field

mxyns:
### Short description
The current implementation of bmp_daemon_set_pd duplicates the "rd" field of bmp_chars as a new field "pd".
This is a priori unnecessary since the rd coming from bgp is not stored in `bmp_chars` but rather in the bgp attributes of the route. Additionally, this duplication may consume a lot of memory (8 bytes * number of routes) as bmp_chars are not deduped using interning.

This PR aims at fixing that by keeping a single field, now internally called "pd" to match the BMP protocol, and exports this field either as "pd" or "rd" in the json outputs depending on the bmp_daemon_set_pd config knob.

I could not test in depth that this implementation is fully sane since I do not have a pcap with bmp and bgp having different RDs/PDs. Maybe @MyLi2tlePony or @GeorgyKirichenko could help with that?
I can however confirm that the BMP field is correctly renamed when bmp_daemon_set_pd=true and that BGP RDs are still exported correctly when BMP does not have a peer distinguisher, regardless of the config knob (BGP takes precedence over BMP for RDs).

### Checklist

I have:

### DIFF
--- a/src/bmp/bmp.h
+++ b/src/bmp/bmp.h
@@ -89,7 +89,6 @@ struct bmp_chars {
   u_int8_t is_out;
   u_int8_t is_loc;
   u_int8_t rib_type; /* new replacing is_in, is_out, is_loc, is_post */
-  rd_t rd;
   rd_t pd;
 
   /* non-key */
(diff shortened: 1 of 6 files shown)


mxyns:
seems to be working fine imo?

Issue #902, opened 2025-11-29
Crash on BGP OPEN


WRMSRwasTaken:
Latest master commit (https://github.com/pmacct/pmacct/commit/2a32635622d805e3d4df7b5628467dea12283f5d) is crashing for about ~50% of the time for me when BIRD opens the BGP session to pmacctd and sending the complete IPv6 DFZ:

gdb bt:
```
Core was generated by `pmacctd: Core Process [default]             '.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000cf089d5b904 in bgp_ls_srcdst_lookup (type=7, pptrs=0x7c6521a95c00) at bgp/bgp_ls.c:1348
1348	        pptrs->igp_src_info = (char *) result->info;
[Current thread is 1 (Thread 0x63cf7b6d7b40 (LWP 460372))]
(gdb) bt
#0  0x00000cf089d5b904 in bgp_ls_srcdst_lookup (type=7, pptrs=0x7c6521a95c00) at bgp/bgp_ls.c:1348
#1  pm_pcap_cb (user=0x7c6521a960a0 "\240_\251!e|", pkthdr=<optimized out>, buf=<optimized out>) at /usr/src/debug/pmacct/pmacct/src/nl.c:152
#2  0x00000cf089cf769f in main (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at /usr/src/debug/pmacct/pmacct/src/pmacctd.c:1394
```

build flags:
```
PLATFORM ..... : x86_64
OS ........... : Linux 6.17.8-hardened2-1-hardened (4f631146fda7)
COMPILER ..... : gcc
CFLAGS ....... : -O2 -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection         -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/woodpecker/src/git.mcl.gg/archmin/archmin/src=/usr/src/debug/pmacct -flto=auto -I/woodpecker/src/git.mcl.gg/archmin/archmin/src/pmacct/src/external_libs/rootfs/include  -I/usr/include   -lpcap  -I/woodpecker/src/git.mcl.gg/archmin/archmin/src/pmacct/src
[...]

PR #905, opened 2025-12-04
initialize result to NULL

mxyns:
### Short description
Hey Paolo!
This result pointer is not initialized to NULL. When the bgp lookup exits prematurely because of an early return, the pointer is not always set to NULL by the `bgp_node_match` function called. This result pointer may then have any value that lingered there on the stack. I had some issues where sfacctd would crash because this result was 0x06. Since it was not NULL, it would pass the `if (result)` check line 1322 and then segfault on access `result->info`. Sometimes this was also fine based on previous calls, for instance if you print the address to debug before that, nothing crashes for some reason :smile: 

Cheers
Max

### Checklist

I have:

### DIFF
--- a/src/bgp/bgp_ls.c
+++ b/src/bgp/bgp_ls.c
@@ -1305,7 +1305,7 @@ void bgp_ls_srcdst_lookup(struct packet_ptrs *pptrs, int type)
   struct in6_addr pref6;
   safi_t safi = SAFI_UNICAST;
 
-  struct bgp_node *result;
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thank you as always Maxence ( @mxyns ), ironically i was looking into the same in order to address Issue #902 and proposed Marcel to exactly try fixing this pointer first. Glad this is all coming to a resolution. Paolo

mxyns:
Haha I should've checked the open issues 🙃 happy to help!

PR #906, opened 2025-12-17
Adding support for custom BMP/BGP packet parsing at runtime

Pedro-W21:
### Short description

This pull request :
- implements a system to load a dynamic library file and extract arbitrarily named functions from it
- implements a new configuration key (`custom_packet_parsing_lib`) to load a dynamic library to replace 4 BGP/BMP functions at runtime in pmacct
- changes how the 4 BGP/BMP functions are called to make it possible to replace them
- adds an option to the docker images build script to copy a dynamic library into the image for use in testing 
- modifies build configuration and install to make it easier to build with pmacct as a dependency

for more details, check the linked PDF explaining this addition more clearly : 
[runtime function replacement for pmacct.pdf](https://github.com/user-attachments/files/24214836/runtime.function.replacement.for.pmacct.pdf)

### Checklist

I have:


Pedro-W21:
The "build system changes" described in the first presentation PDF are no longer entirely correct, to pass all checks, the following has changed :

there is no longer a build_config.h file generated upon compiling configuration to `#define` all global symbols to their configured values (e.g. `WITH_GNUTLS`)

the previous version of the code (with build_config.h) had some symbols which were generated automatically by autotools, like `VERSION` or `PACKAGE_NAME`

Those standard symbols were also defined within ndpi, which led to symbol redefinition warnings at compilation in files where both build_config.h and ndpi were included

while this generated code that worked at runtime for the test-framework suite, the CI/CD checks include compilation on all platforms with the `--enable-debug` option, which turns warnings into errors (via `-Werror`), so those checks fail

Since there is no intended way to avoid outputting those standardised symbols into a build header file like build_config.h when one exists, to fix the issue while still allowing building with pmacct as a dependency, the build header file was removed in favor of including all definitions of relevant symbols into the `CFLAGS` exported in the `pmacct.pc` file that this pull request adds

this was done by introducing a new function and variable into the `configure.ac` file :
- `PKG_CONFIG_DEFS` is the variable containing all definitions passed to the compiler that should be exported to `pmacct.pc`
- `PKG_AC_DEFINE_PUBLIC(name, value, description)` bundles an `AC_DEFINE` operation and adding a definition CFLAG to `PKG_CONFIG_DEFS`

All relevant calls to `AC_DEFINE` within `configure.ac` have been changed to that new function in order to cleanly export those definitions for programs compiling with pmacct as a dependency.

With this new system, any new symbols that need to be made public for correct compilation should be defined using `PKG_AC_DEFINE_PUBLIC` to avoid potentially breaking programs depending on pmacct.

docker/base/Dockerfile
```diff
@@ -54,6 +54,8 @@ ENV MAKEFLAGS="-j${NUM_WORKERS}"
 # Do not check certificates in wget for external deps
 ARG DEPS_DONT_CHECK_CERTIFICATE
 RUN ./ci/deps.sh
+
```
msune: nit: remove

docker/base/Dockerfile
```diff
@@ -90,4 +93,10 @@ RUN apt-get update && \
   apt-get -y clean && \
   rm -rf /var/lib/apt/lists/* && \
   ldconfig
+
+
+ARG CUSTOM_LIB_PATH=ci/libnothing.so*
```
msune: I don't understand this. Why does the base container need an ARG, and why does it point to `ci/libnothing.so*`?

I think this and the next line needs to be removed

src/bgp/bgp.c
```diff
@@ -27,6 +27,7 @@
 #include "bgp_blackhole.h"
 #include "bgp_ls.h"
 #include "thread_pool.h"
+#include "custom_packet_parsing/packet_processor.h"
```
msune: I would suggest to rename the folder `custom_packet_parsing` to `packet_processor`, following the naming used

src/custom_packet_parsing/dynamic_loading.c
```diff
@@ -0,0 +1,55 @@
...
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <dlfcn.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "pmacct.h"
+#include "dynamic_loading.h"
+
+bool dynlib_fn_end(const struct dynlib_fn* fn) {
```
msune: Remove this function: `dynlib_fn_end(fn)` is longer than the body itself (`!fn->name`)

At the same time, are you sure this is the right way to check for the las symbol?

src/custom_packet_parsing/dynamic_loading.c
```diff
@@ -0,0 +1,55 @@
...
+  void* handle = dlopen(lib.path, RTLD_NOW);
+  // load the dynamic library from name/path
+  if (!handle) {
+    Log(LOG_ERR, "ERROR ( %s ): [dynlib] Could not load provided library %s: %s\n", config.name, lib.path,
+        dlerror());
+    return DL_Error;
+  }
+
+  struct dynlib_fn* fn;
+  for (int index = 0; !dynlib_fn_end(fn = &lib.table[index]); index++) {
```
msune: This:

> ` !dynlib_fn_end(fn = &lib.table[index])`

is difficult to read and likely problematic:

1. See comment on `dynlib_fn_end()`.
2. For safety you should check that table is not NULL.
3. I would suggest a more compact (note, indentation with spaces in comments, as I can use hard tabs, so don't copy&paste):

```
struct dynlib_fn* fn;
int index = 0;
do {
    if (!lib.table)
       return DL_Error; //See comment ont return codes.     

    fn = lib.table[index];
    if (!fn->name)
       break;
    if (!fn->store)
       continue;

    // ....
} while(true);
```

More comments on the header file.

src/custom_packet_parsing/packet_processor.h
```diff
@@ -0,0 +1,48 @@
...
+#include "bmp/bmp.h"
+#include <time.h>
+
+/// Packet processor, set of functions called to parse a received bmp/bgp buffer
+typedef struct {
+    int (* bgp_parse_msg) (struct bgp_peer *peer, time_t time, int online);
+    u_int32_t (* bmp_process_packet) (char * buf, u_int32_t buf_len, struct bmp_peer *peer, int *term);
+    int (* bmp_peer_init) (struct bmp_peer *peer, int func_type);
+    void (* bmp_peer_close) (struct bmp_peer *peer, int func_type);
+} packet_processor;
```
msune: nit: define it as:

```
typedef struct packet_processor {

} packet_processor_t;
```

So that you can use `struct packet_processor` in a forward declaration without including the header. It also makes it clear `packet_process_t` is a typedef.

src/custom_packet_parsing/dynamic_loading.h
```diff
@@ -0,0 +1,49 @@
...
+};
+
+enum dynlib_result {
+  DL_Error,
+  DL_Success,
+};
+
+/// Load a dynamic library, retrieve symbols and store them where @dynlib_fn#store points to
+/// see @dynlib for more details on the definition of a library
+extern enum dynlib_result dynlib_load_and_resolve(const struct dynlib lib);
```
msune: Is all of this header really needed?

Wouldn't it be sufficient to do:

```
//Forward declare
struct packet_processor;

/**
 *  Dynamically load a custom packet processor from an object file
 */
int packet_processor_dynload(const char *lib_path, struct packet_processor *processor);
```

Where processor is pre-filled with the defaults? I don't see the need to define this intermediate `struct dynlib`, as the hook points are anyhow fixed and always the same ones.

src/custom_packet_parsing/packet_processor.h
```diff
@@ -0,0 +1,48 @@
...
+
+/// The in-use packet processor. Should not change after config has been loaded.
+/// Defaults to @DEFAULT_PACKET_PROCESSOR
+extern packet_processor active_packet_processor;
```
msune: There can only be one packet_processor at runtime, therefore simply remove `active_packet_processor` and `DEFAULT_PACKET_PROCESSOR`, and do something like:

```
extern packet_processor pprocessor;
```

src/custom_packet_parsing/packet_processor.h
```diff
@@ -0,0 +1,48 @@
...
+const static packet_processor DEFAULT_PACKET_PROCESSOR = {
+    bgp_parse_msg,
+    bmp_process_packet,
+    bmp_peer_init,
+    bmp_peer_close
+  };
+
+/// Use dynlib to load the parsing library located at "lib_path"
+/// "lib_path" can be libname.so, or a relative/absolute path (see man @dlopen)
+extern enum dynlib_result load_parsing_lib(const char* lib_path);
```
msune: I think this should be:

```
int packet_processor_dynload(const char *lib_path, struct packet_processor *processor);
```

test-framework/tools/pmacct_build/build_docker_images.sh
```diff
@@ -1,12 +1,34 @@
 #!/bin/bash
 
+
 SCRIPT_DIR=$( cd -- "$( dirname -- "$0" )" &> /dev/null && pwd )
 PMACCT_ROOT_LOCATION="$SCRIPT_DIR/../../../"
 
 TAG='_build'
 
+CUSTOM_LIB_PATH="ci/libnothing.so*"
```
msune: Idem comment above.

configure.ac
```diff
@@ -12,6 +12,14 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES(yes)])
 PMACCT_VERSION=`cat $srcdir/VERSION`
 AC_SUBST(PMACCT_VERSION)
 
+AC_DEFUN([PKG_AC_DEFINE_PUBLIC], [
```
msune: I understand the aim here is to capture all the `AC_DEFINES`, so that when buliding packet processors you have the same (AC) defines.

But I don't think this is the right approach. It bloats all configure.ac (and the `$3` is filled up with garbage). Please revert this change.

Instead, create a distributable header file with all the MACROS defined, by using the `AC_CONDITIONAL` statements here: https://github.com/pmacct/pmacct/blob/master/configure.ac#L1253-L1277

configure.ac
```diff
@@ -1276,6 +1284,12 @@ AM_CONDITIONAL([USING_BGP_BINS], [test x"$USING_BGP_BINS" = x"yes"])
 AM_CONDITIONAL([USING_BMP_BINS], [test x"$USING_BMP_BINS" = x"yes"])
 AM_CONDITIONAL([USING_ST_BINS], [test x"$USING_ST_BINS" = x"yes"])
 
+A_PC_CFLAGS="$CFLAGS $PMACCT_CFLAGS"
```
msune: nit: `A_PC_` to `PMACCT_PC_`

Makefile.am
```diff
@@ -7,7 +7,8 @@ EXTRA_DIST = include sql examples docs CONFIG-KEYS FAQS QUICKSTART UPGRADE
 if USING_SQL
 EXTRA_DIST += src/sql_common_m.c
 endif
-
```
msune: nit: leave blank

msune (changes requested):
@Pedro-W21 sorry for the late review on this PR.

Overall, kudos for the work. I think this PR needs some work, but it looks promising.

My suggested plan of action:

1. Look into the comments. I think things can be simplified quite a bit.
1. Squash commits, so that review is easier and it's more understandable (right now ~38 commits), and force push. One possible split could be:
   * packet processor struct definition (header file only)
   * packet processor `dlopen` functionality (only filling up the struct, nothing more)
   * Extend pmacct configuration routine and config docs
   * Changes in the build system, including docker
   * README
1. I think this PR needs CI coverage. Some initial thought: create a dummy Rust packet processor that outputs some trace and calls back to the pmacct C function, and then run pmacct with BMP to see those traces. 

Good job!

docker/base/Dockerfile
Pedro-W21: > Why do we need to copy that in first place in the docker image? Shouldn't pmacct boot normally unless you define the custom config key?

Pmacct does boot normally unless the config key is used, the copy is only needed when test configurations are modified to use a packet processor with the new config key

> Simply use [docker volumes](https://docs.docker.com/engine/storage/volumes/) mounting the test packet_processor along with pointing the config to the right path where the volume is mounted in the test.

would the following changes to the test system to implement that be ok ?

- in `test-framework/library/sh/pmacct_docker/docker-compose-template.yml`
    - add a line with a marker to replace e.g.
    ```
    volumes :
        [...]
        - ${PACKET_PROCESSOR}
    ```
- in `test-framework/library/py/test_params.py`
    - add a new field to KModuleParams : `packet_processor` initialised to None by default from a named param in the constructor (`__init__` method)
    - if the param is not None, initialize the field to  `self.test_folder + packet_processor`
- in `test-framework/library/py/setup_test.py`, in the `create_pmacct_compose_files` function :
    - if the new field in the params is None, remove the volume tagged with the marker from the `vols` list
    - if it's a path, replace the volume marker with `  - {params.packet_processor}:/usr/local/lib/libparse.so`
- in `test-framework/library/py/setup_test.py` in the `edit_config_with_framework_params` function :
    - if the new field is None, don't change configuration
    - if it's a path, update the `packet_processor` key in the configuration to `/usr/local/lib/libparse.so`
    
This should leave default behavior unchanged while allowing the mounting of a given library into the container of a test that needs it

and of course the current implementation copying the library onto the image would be removed

src/bgp/bgp.c
mxyns: Agreed

src/custom_packet_parsing/dynamic_loading.h
mxyns: We want to have this dynlib thing available for other uses as well, hence the generic implementation. We can move dynlib into its own directory to make this intent clearer. What do you think?

src/custom_packet_parsing/packet_processor.h
mxyns: Im sorry I dont get this one, wouldn't that result in just renaming the global variable?

mxyns: Im sorry I dont get this one, wouldn't that result in just renaming the global variable?

src/custom_packet_parsing/dynamic_loading.h
msune: > We want to have this dynlib thing available for other uses as well, hence the generic implementation

No, I don't think so. 

Let's please apply the [KISS](https://en.wikipedia.org/wiki/KISS_principle) principle. If we ever have someone else trying to dynamically load code in pmacct, we can evaluate whether to create an abstraction to load symbols, as we will better understand both use-cases.

Having said that, even if that were the case, the current implementation of `dynlib_load_and_resolve()` provides no real value. It forces you to define  `struct dynlib` and use a dynamic array to load , which in turn requires a new header. The inputs of that array are always fixed (the symbols you are trying to load) as they strictly depend on the code given that you need to define the hook points, and modify them with the indirect call.

So this function is really not useful. It's way easier to define something along the lines of:

```
 1                                                                                 
  2 //Typedef so casting is easier when filling up the                              
  3 //packet_processor struct                                                       
  4 typedef int (* bgp_parse_msg_f) (struct bgp_peer *peer, time_t time, int online);
  5 //... add the rest                                                              
  6                                                                                 
  7 struct {                                                                        
  8         bgp_parse_msg_f parse_msg;                                              
  9         //...                                                                   
 10 } packet_processor;                                                             


  // Note this assumes processor has been prefilled during extern initialization with default
  // implementation function ptrs.
                                                                                
 12 int packet_processor_dynload(const char *lib_path, struct packet_processor *processor) {
 13                                                                                 
 14         const char *err_symbol = NULL;                                          
 15                                                                                 
 16         //dlopen etc.                                                           
 17         //...                                                                   
 18                                                                                 
 19         processor->parse_msg = (bgp_parse_msg_f) dlsym(handle, "bgp_parse_msg_hook");
 20         if (!processor->parse_msg) {                                            
 21                 err_sym = "bgp_parse_msg_hook";                                 
 22                 goto ERR;                                                       
 23         }                                                                       
 24                                                                                 
 25         //...                                                                   
 26                                                                                 
 27 ERR:                                                                            
 28         printf("ERROR: expected symbol \"%s\" not found in dynamic library %s.\n",
 29                 err_sym, lib.path);                                             
 30         exit(1);                                                                
 31 }         
```

Note that I added `_hook` at the end, to make more explicit that it's a different symbol than the one defined in pmacct, but with `bgp_parse_msg_f` signature.

(And in case you are concerned of the use of `goto ERR`, as opposed to the general use of it, this pattern is a well known pattern to handle exit or error conditions, see: https://elixir.bootlin.com/linux/v6.14.5/source/Documentation/process/coding-style.rst section 7.)

docker/base/Dockerfile
msune: My main concern is not to have any test / dummy custom packet processor implementation as part of the base container or any other pmacct container. As long as this is respected, I think it's fine.

I can't comment on the `test-framework/` itself. I already provided my input to the design of that test framework ([here](https://github.com/pmacct/pmacct/pull/774#issuecomment-2074517338), and I continue to think - and has been shown several times, IMO, with multiple regressions - that this approach is not the right one.

I would very much prefer a good scoped smoke test as per described:

> I would strongly suggest to use the approach of https://github.com/pmacct/pmacct/pull/885 for tests instead.

where pmacct is started and we verify that the hooks are executed and we strictly control inputs and outputs to the test. If you want guidance on that, we can discuss it.

src/custom_packet_parsing/packet_processor.c
```diff
@@ -0,0 +1,42 @@
...
+        {"bmp_process_packet", (void**) &processor.bmp_process_packet},
+        {"bmp_peer_init", (void**) &processor.bmp_peer_init},
+        {"bmp_peer_close", (void**) &processor.bmp_peer_close},
+        {NULL,NULL}
+      },
+    }
+  );
+
+  // On dynlib success, actually set active packet processor to the loaded library
+  if (result == DL_Success) {
```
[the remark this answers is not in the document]
mxyns: it does. see the use in `cfg_custom_packet_parsing_lib`. the log message there is wrong though

https://github.com/pmacct/pmacct/pull/906/files/4ffd08aa531201be5b83426e45b85e1722c397f2#diff-6513d050521a59fe3d218bfe016f545ab1cf747e40169b6b1c74177ad1863bcdR8998-R9004

i can't include a code snippet for some reason, sorry.

src/custom_packet_parsing/packet_processor.h
mxyns: I feel like this naming is less clear though out of the header file, especially where packet processor callbacks are called.

```c
      if (ret <= 0) {
        Log(LOG_INFO, "INFO ( %s/%s ): [%s] BMP connection reset by peer (%d).\n", config.name, bmp_misc_db->log_str, peer->addr_str, errno);
        FD_CLR(peer->fd, &bkp_read_descs);
        pprocessor.bmp_peer_close(bmpp, FUNC_TYPE_BMP); // first time reading this line: "what is a pprocessor?"
        recalc_fds = TRUE;
        goto select_again;
      }
```

that might be subjective

src/custom_packet_parsing/dynamic_loading.c
mxyns: i moved to a simple for loop, not sure how i did not write it like this initially

mxyns:
Hi @msune

I tried to address most of your comments, including the squash of commits to cleanup everything. I might have missed some because the Github GUI is horrible when tracking multiple threads of comments. 

Everything builds locally and the flags are passed correctly to the parsing library, as far as I could tell. The CI fails just like the master does atm because of python3 installation in the image so I'll check in the following days if I can rerun the CI.

Could you please have another look at the PR whenever? It should be easier to review with everything squashed.

Thanks a lot for your time

PR #912, opened 2026-03-03
export bgp id in bmp route monitor

mxyns:
### Short description
the bgp id is exported is all messages featuring the per-peer header excepted the route-monitor message
i believe this is due to the fact that the bgp_id was stored in the `struct bmp_data` and that the only value accessible from the msglog function for bmp route monitor messages was the `struct bmp_chars` which does not feature the bgp_id.

this PR tries to allow exporting the bgp_id field in the route_monitor message. it does so by moving the bgp_id field from the `struct bmp_data` to the `struct bmp_chars`, in the non-key section of the struct to avoid impacting the bgp processing logic. it is now exported as "bgp_id". i did not update the bgp/bmp fields documentation, as, to my understanding, it was already hinting that the bgp_id was exported in bmp, but I'm not sure I understood it correctly.

i also moved the type of the bgp_id, **in bmp only**, from `struct host_addr` to `struct in_addr` since, afaik, the bgp id's only valid format the 4 bytes represented as an ipv4 address. this saves a small 12 bytes per bgp_id. I believe this could also be modified in bgp to gain a bit of memory space, unless I missed some pmacct specific behaviour related to that bgp_id that requires it to be the size of an ipv6?

### Checklist

I have:
- [~] compiled & tested this code **in my lab** i'll let the ci run the test suite and see what happens


paololucente (commented):
Thanks for this contribution Maxence ( @mxyns ), much appreciated it as always. I have tested it working on my setup, merging. Paolo

PR #918, opened 2026-04-24
* BMP daemon: add EVPN route-monitor parsing and JSON fields

jaredmauch:
### Short description
Wires in the BMP code similar to the BGP code for emitting EVPVN routes

### Checklist
I have:
- [ x ] compiled & tested this code
- [ x] included documentation (including possible behaviour changes)


paololucente (commented):
Hey Jared ( @jaredmauch ),

Thanks very much for this monumental contribution, support for EVPN in BGP and BMP is something that has been long in the radar as a future feature, so this all goes in the right direction.

Could you by any chance share (here or privately) a BMP trace that can be replayed with bmp_play ( https://github.com/paololucente/bmp_play ) so that i can test the code at my end? This would allow me to take care of a couple of improvement to this PR, for example:

* Dynamic and conditional allocation of EVPN data
* Better integrate code in bgp_nlri_parse() to avoid the `continue`
* Move code in a new bgp_evpn.c file

I would have also a question for you since right now only msglog is supported. Would we store EVPN data in memory, effectively building a RIB, for supporting dumps at regular time intervals later: how would you propose to organize it? As a RIB with longest-prefix match using maybe `evpn_ip` for that or rather as a hash where we define a key (and if so which one?) so that entries can be inserted / updated for non-key attributes / removed? What would be useful?

Paolo

jaredmauch:
> Hey Jared ( @jaredmauch ),
> 
> Thanks very much for this monumental contribution, support for EVPN in BGP and BMP is something that has been long in the radar as a future feature, so this all goes in the right direction.
> 
> Could you by any chance share (here or privately) a BMP trace that can be replayed with bmp_play ( https://github.com/paololucente/bmp_play ) so that i can test the code at my end? 

So I just pushed up some anonymized data for the test suite and added the 209 test, is that enough to satisfy what you need on this?  if not, I can work to get you the private data which is more complex.

PR #920, opened 2026-04-28
add ecommunity format raw

MyLi2tlePony:
### Short description

Added bgp_ecommunities_raw flag to allow Extended Communities to be output in raw type:subtype:value1:value2 format instead of the default human-readable representation. This is useful for debugging and for preserving visibility into unknown or unsupported Extended Community types.

### Checklist

I have:

### DIFF
--- a/CONFIG-KEYS
+++ b/CONFIG-KEYS
@@ -1759,19 +1759,8 @@ DEFAULT:        false
 
 KEY:		bgp_comms_num
 DESC:		Defines whether known Standard (ie. "internet") and Extended Communities (ie. "RT")
-		shdould be presented in string or numerical format. By default communities are looked up
+		should be presented in string or numerical format. By default communities are looked up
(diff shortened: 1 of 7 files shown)


paololucente (commented):
Hi Andrey ( @MyLi2tlePony ),

Thanks for this much wanted contribution! Code makes total sense. I would have one ask for you: in this PR you define a new `bgp_ecommunities_raw`; while reviewing existing code i came to the realization that `bgp_comms_num` does its job for standard communities but was very poorly ported to extended communities. I was thinking to basically re-use that config knob for your code. Would you also believe that does make sense?

Paolo

MyLi2tlePony:
Hi [Paolo](https://github.com/paololucente),

Thanks for the review and suggestion — that makes sense to me. I’ve updated the PR to re-use the existing bgp_comms_num knob instead of introducing bgp_ecommunities_raw.

When you have a chance, could you please take another look?

Thanks!

PR #926, opened 2026-07-03
fix(bgp): fix memory leaks in BGP-LS and UPDATE attribute parsing

riccardo-negri:
### Short description
- bgp_parse_update_msg: unintern partially-parsed attr components (aspath, community, ecommunity, lcommunity) when bgp_attr_parse() returns an error, preventing leaks on malformed UPDATE messages

- bgp_attr_parse: use goto cleanup label so as4_path is always uninterned on all early-return error paths inside the parse loop

- bgp_ls_nlri_parse: initialize tlv_type to 0 to prevent undefined behaviour when the inner TLV loop does not execute, which could cause phantom RIB insertions

- bgp_ls_nlri_parse: free attr_hdr and attr_aux when cdada_map_insert_replace fails to avoid leaking both allocations

### Checklist
I have:


paololucente (commented):
Ciao Riccardo ( @riccardo-negri ), thanks for catching this. I have tested it successfully at my end too. Merging. Paolo

PR #928, opened 2026-07-06
fix(nfacctd): replace stale vrf_name_map entry on VRF ID reassignment

riccardo-negri:
### Short description
cdada_map_insert() silently ignores updates to existing keys (returns CDADA_E_EXISTS) leaving the old mapping in place permanently.

When a router reassigns its internal VRF IDs, both the old and new mappings coexist in vrf_name_map indefinitely. Consider the following before/after state on the router:

Before:  ID_1 -> NAME_A,  ID_X -> NAME_B
After:   ID_1 -> NAME_B,  ID_2 -> NAME_A

The router sends both new option records. ID_2 -> NAME_A is a fresh key and inserts correctly. ID_1 -> NAME_B hits an existing key and the update is silently dropped. nfacctd then permanently holds:

ID_1 -> NAME_A  (stale — should be NAME_B)
ID_2 -> NAME_A  (correct)

All flows arriving with ID_1 are enriched with NAME_A instead of NAME_B, causing flows from one customer to appear under another customer's VRF context in the output. The daemon retains this wrong state until restarted, which explains why a restart temporarily resolves the issue.

Replace cdada_map_insert() with cdada_map_insert_replace(), which atomically updates existing keys and returns the old pointer for freeing, handling both insert and update in a single call.

### Checklist
I have:

### DIFF
--- a/src/nfacctd.c
+++ b/src/nfacctd.c
@@ -2353,10 +2353,16 @@ void process_v9_packet(unsigned char *pkt, u_int16_t len, struct packet_ptrs_vec
 	      memset(vrf_name, 0, MAX_VRF_NAME_STR_LEN);
 	      memcpy(vrf_name, (const char *)pkt + tpl->fld[NF9_VRF_NAME].off[0], MIN(tpl->fld[NF9_VRF_NAME].len[0], (MAX_VRF_NAME - 1)));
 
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thanks Riccardo ( @riccardo-negri ), makes sense! Thanks for catching this! Merging! Paolo

PR #929, opened 2026-07-09
fix(nfacctd): replace stale in_rd_map, out_rd_map and iface_name_map entries on ID reassignment

riccardo-negri:
### Short description
Same issue as 7b8ef26: cdada_map_insert() silently ignores updates to existing keys (returns CDADA_E_EXISTS), leaving stale mappings in place permanently.

Replace cdada_map_insert() with cdada_map_insert_replace() for in_rd_map, out_rd_map and iface_name_map, freeing any displaced entry. Each RD map now owns an independent heap-allocated copy of the RD value so replacements on one map cannot dangle the other. The intermediate mpls_vpn_rd heap allocation is replaced with a stack variable since it is only needed as a temporary parse buffer.

### Checklist
I have:

### DIFF
--- a/src/nfacctd.c
+++ b/src/nfacctd.c
@@ -2267,7 +2267,7 @@ void process_v9_packet(unsigned char *pkt, u_int16_t len, struct packet_ptrs_vec
 
 	  if (entry) {
 	    u_int32_t ingress_vrfid, egress_vrfid;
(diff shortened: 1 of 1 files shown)


paololucente (commented):
Thanks again Riccardo ( @riccardo-negri ) also for this fix. Merging. Paolo

PR #932, opened 2026-07-13
add bmp peer up local as

MyLi2tlePony:
### Short description

Adds the local BGP AS number (local_as) to BMP Peer Up Notification (message type 3) output, exposing the AS carried in the local BGP OPEN message alongside the existing peer_asn.

### Checklist

I have:


paololucente (commented):
Thanks Andrey ( @MyLi2tlePony ),

useful improvement as always! Thanks for also fixing the tests & merging. 

Paolo

PR #935, opened 2026-07-25
fix, JSON/Avro plugins: encode tcp_flags as an integer instead of a string

ssivakorn:
### Short description

**tcp_flags** is emitted as a string in both the JSON and Avro output plugins, instead of an integer/long, with no apparent reason. tcp_flags is a ``u_int8_t`` bitmask (e.g. SYN+ACK = 18) — not a symbolic string — so this forces every consumer to parse a stringified number back into an integer for no benefit.

This appears to be a long-standing inconsistency rather than an intentional design choice:

- Neighboring numeric primitives in the exact same functions (``ip_tos``, ``sampling_rate``) are already correctly emitted as json_integer/avro_schema_long.
- The formatted/CSV print plugin (``print_plugin.c``) and the SQL output (``sql_common.c``) already treat tcp_flags as a plain numeric value (%u, unquoted).
- Only the JSON and Avro composers stringify it, via ``sprintf(misc_str, "%u", tcp_flags)`` followed by ``json_string()``/``avro_value_set_string()``.
- Traced the git history back to when JSON/Avro output for tcp_flags was first introduced (~2016-2017) — no commit message, comment, or existing issue documents a rationale for the string type.

This PR changes tcp_flags to be emitted as json_integer/avro_schema_long (and avro_value_set_long) in the default (non-array) encoding path, matching how ip_tos/sampling_rate are already handled. The opt-in ``tcpflags_encode_as_array`` encoding (symbolic flag names) is untouched.

### Checklist

I have:
  - [x] compiled & tested this code (built pmacct with this patch applied, ran nfacctd, and separately verified the exact compose_json_tcp_flags function against a real jansson build before/after: {"tcp_flags":"18"} → {"tcp_flags":18})
  - [x] added the LICENSE template to new files (n/a — no new files)
  - [ ] included documentation (behavior change noted above; let me know if a docs update is expected elsewhere, e.g. CONFIG-KEYS)

### DIFF
--- a/src/plugin_cmn_avro.c
+++ b/src/plugin_cmn_avro.c
@@ -327,7 +327,7 @@ avro_schema_t p_avro_schema_build_acct_data(u_int64_t wtc, u_int64_t wtc_2, u_in
       compose_tcpflags_avro_schema(schema);
     }
     else {
-      avro_schema_record_field_append(schema, "tcp_flags", avro_schema_string());
+      avro_schema_record_field_append(schema, "tcp_flags", avro_schema_long());
     }
   }
 
@@ -1068,14 +1068,12 @@ avro_value_t compose_avro_acct_data(u_int64_t wtc, u_int64_t wtc_2, u_int64_t wt
 #endif
 
   if (wtc & COUNT_TCPFLAGS) {
-    sprintf(misc_str, "%u", tcp_flags);
-
     if (config.tcpflags_encode_as_array) {
       compose_tcpflags_avro_data(tcp_flags, value);
     }
     else {
       pm_avro_check(avro_value_get_by_name(&value, "tcp_flags", &field, NULL));
-      pm_avro_check(avro_value_set_string(&field, misc_str));
+      pm_avro_check(avro_value_set_long(&field, tcp_flags));
     }
   }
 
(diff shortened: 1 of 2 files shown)


ssivakorn:
+ @sagottlieb @awright

paololucente (commented):
Dear Suphannee ( @ssivakorn ), the code change makes sense. Can you please make it against master code rather than 1.7.9? Unfortunately CI/regression, due to old packages used by 1.7.9 and now deprecated, does not work anymore. Paolo

ssivakorn:
> Dear Suphannee ( @ssivakorn ), the code change makes sense. Can you please make it against master code rather than 1.7.9? Unfortunately CI/regression, due to old packages used by 1.7.9 and now deprecated, does not work anymore. Paolo

Will do! Thank you for reviewing @paololucente 😄

ssivakorn:
> > Dear Suphannee ( @ssivakorn ), the code change makes sense. Can you please make it against master code rather than 1.7.9? Unfortunately CI/regression, due to old packages used by 1.7.9 and now deprecated, does not work anymore. Paolo
> 
> Will do! Thank you for reviewing @paololucente 😄

Re-opened here: https://github.com/pmacct/pmacct/pull/937.

Thank you @paololucente

PR #936, opened 2026-07-26
Allow to configure syslog's ident

evertonen:
Some openlog() implementations like musl assume an empty string instead of the program name. This patch will not break backwards compatibility for cases where a symlink is used to change the program name.

- [X] included documentation


paololucente (commented):
Thanks Everton ( @evertonen ), this is an useful contribution. merging. Paolo

